PR #16, opened 2018-03-13
Register with JSAPI as named consumer

andytuba:
And use named-consumer-specific containers

e.g.
```html
<a href="/user/whatever">
<span>
    <!-- this is the main jsapi container -->
    <span data-name="toolbox">
        <!-- this is the specific part of the container requested by toolbox -->
    </span>
</span>
```

screen shot 2018-03-13 at 2 31 09 pm


creesch:
Awesome, works like a charm!

PR #121, opened 2019-07-17
modtab-threshold event handler now uses input event; 

adzm:
this responds to changing the value of numeric input via the mouse or accessibility tools using scrollbars. Previously it used keypress which would not respond to those. Using change would only react upon blur of the input element. The input event is supported by all modern browsers, IE9+

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -639,10 +639,10 @@ function queuetools () {
             });
 
             // Set reports threshold (hide reports with less than X reports)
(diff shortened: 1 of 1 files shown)


adzm:
Browser name:	Chrome 
Browser version:	Chrome 76
Module:	Queue Tools
Issue location:	Modqueue
Issue description:	Changing the report filter threshold only applies when using the keyboard; changing with the mouse or accessibility tools do not trigger the keypress event.

creesch:
I see no immediate faults with this and it looks simple enough. I'll merge it.

PR #126, opened 2019-08-02
Modqueue fixes for never-ending reddit

adzm:
Details in commits. The $things variable was captured by some event handlers and is not updated when NER loads more items.

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -561,7 +561,7 @@ function queuetools () {
             });
 
             $('.unhide-selected').click(() => {
-                $things.show();
+                $('.thing').show();
             });
 
             // Expand reports on click.
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -360,7 +360,7 @@ function queuetools () {
             <a href="javascript:;" class="pretty-button action neutral"  accesskey="R" type="neutral"  tabindex="4">remove&nbsp;selected</a>
             <a href="javascript:;" class="pretty-button action positive" accesskey="A" type="positive" tabindex="5">approve&nbsp;selected</a>
         </span>
-        ${viewingspam ? '' : `<span><a><label for="modtab-threshold">Report threshold: </label><input id="modtab-threshold" type="number" value="${reportsThreshold}" /></a></span>`}
+        ${viewingspam ? '' : `<span><a><label for="modtab-threshold">Report threshold: </label><input id="modtab-threshold" type="number" min="0" value="${reportsThreshold}" /></a></span>`}
         <span class="dropdown-title lightdrop" style="float:right"> sort:
             <div class="tb-dropdown lightdrop">
                 <span class="selected sortorder">${listingOrder}</span>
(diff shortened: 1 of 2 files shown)


extension/data/modules/queuetools.js
```diff
@@ -561,7 +561,7 @@ function queuetools () {
             });
 
             $('.unhide-selected').click(() => {
-                $things.show();
+                $('.thing').show();
```
creesch: As never-ending reddit adds thing to the dom I don't think this will work properly for things added after page load. You probably want to do `$body.find('.thing').show();`

extension/data/modules/queuetools.js
```diff
@@ -649,7 +649,9 @@ function queuetools () {
 
                 $(this).val(threshold);
                 self.setting('reportsThreshold', threshold);
-                setThreshold($things);
+                
+                const things = $('.thing');
```
creesch: Our policy is to start variables that are jquery objects with `$` so it is clear it is a jquery object. Also here goes that you probably want to use `$body.find('.thing')`. 

I saw there already is a `$things` variable declared and can understand you not want to deal with that so this isn't the biggest of deals to me here.

extension/data/modules/queuetools.js
adzm: hm, it seems like `$('.thing')` and `$body.find('.thing')` result in the same output; looks like jquery's selectors will run on body by default. I'll test this out some more though.

Geo1088: `$()` searches the entire document for the given selector, so since there's no `.thing` in `<head>`, it should be fine as-is.

I'm not sure what creesch was talking about here, actually. Using `$things` - which is a predefined set of elements - wouldn't include elements added to the DOM since the value was stored, but since you changed it to `$('.thing')` - which generates a new list of elements - it should work fine the way you have it here.

@creesch was this just you misreading the diff or is there more to it?

creesch: No actually things that get added dynamically to the DOM don't always get selected by jquery but if you use an existing reference element like body then it can find them. It is the better way to select elements that get dynamically added to the DOM.

extension/data/modules/queuetools.js
adzm: I changed this to use 


```
                const $allThings = $('.thing');
                setThreshold($allThings);
```

and rebased the patch so as not to clutter the history

extension/data/modules/queuetools.js
adzm: stepping into the code, appears that jQuery will do this for the selector:

`return ( context || root ).find( selector );`

where `root` is the document element

which then eventually does

`jQuery.find( selector, self[ i ], ret );`

and

`push.apply( results, context.getElementsByClassName( m ) );`

looks like querySelectorAll is used instead of getElementsByClassName if the selector is complicated enough; this is all in the Sizzle function in jquery source.

These things really do seem to be equivalent; using the alternate form just gets the body via selector and then goes into find/Swizzle with body as context explicitly instead of implicitly via the default root (document)

I'm fine with making the change if necessary but I can confirm from testing that this does seem to consistently apply to all matching elements in the DOM.

creesch: Alright, we also discussed it in discord and geo agrees with you. There is something with dynamically added elements and selectors at least with event handlers and I might be mixing stuff up. In any case here it seems indeed to be fine.

PR #148, opened 2019-10-04
feat(BetterButtons): added post sticky position

### LINKED ISSUE #69: Betterbuttons "sticky" button on listing pages
https://github.com/toolbox-team/reddit-moderator-toolbox-legacy/issues/593

> Because of the new double sticky feature on reddit, I think something like this might become super useful.
> 
> [The reddit API](https://www.reddit.com/dev/api#POST_api_set_subreddit_sticky) for stickies has a feature to specify which slot you want the sticky to go in to, which this could take advantage of.
> 
> It could be a new link in the gutter of a listing, like this:
> 
> 
> 
> And when you click it, you see this:
> 
> 
> 
> And an "unstick" link on existing stickies, so you could unstick a sticky from the listing too.

SpyTec:
fixes #69 

This adds positional sticky ability, process looks like pictures below. 1 indicates top position, 2 indicates bottom. I think best would be to change this to top / bottom instead of 1 and 2. Thoughts?

Unsticky has been tested to work as intended with the changes

image

image

image

### DIFF
--- a/extension/data/modules/betterbuttons.js
+++ b/extension/data/modules/betterbuttons.js
@@ -373,9 +373,9 @@ function betterbuttons () {
                     ${!unsticky ? `
                     <span class="tb-sticky-position" style="display: none;">
                         <span class="error close" style="">sticky?</span>
-                        <a class="tb-bracket-button tb-sticky-post" data-sticky-spot="1" href="javascript:;">1</a>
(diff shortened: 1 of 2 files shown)


creesch:
> top / bottom instead of 1 and 2

Yup sound good. Or possibly just do `1. top / 2. bottom`

SpyTec:
Alright, I left it to `top / bottom`. The numbers don't mean anything to users anyway :P

PR #149, opened 2019-10-04
fix(ModBar): Keep "Open in old Reddit" button synced with URL in new Reddit

### LINKED ISSUE #142: "open in old Reddit" button doesn't update its href when navigating different pages on new Reddit
The current code sets the `href` of the link once when the page loads, but since new reddit is a SPA that updates `window.location` without reloading the page, the button still points to the initial page unless the browser is reloaded manually.

Ideally we can still keep the button as a link, because that makes middle-clicking work, and instead just update the `href` when we get a page navigation event from the jsapi.

SpyTec:
fixes #142

### DIFF
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -94,6 +94,8 @@ function modbar () {
                 $footerblock.hide();
             } else {
                 $footerblock.show();
+                const {url} = getDirectingTo();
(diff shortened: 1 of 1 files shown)


SpyTec:
Yeah something broke.

SpyTec:
Synced with upstream

creesch:
I forgot, did you still need to change something in this one?

SpyTec:
Nope!

Issue #94, opened 2019-06-12
Add comment lock functionality to more toolbox functions


creesch:
- [ ] Macros
- [ ] Removal reasons
- [ ] Comment nuke (checkbox to lock everything instead of nuking)

SpyTec:
With comment nuke, should we just modify the popup to have a checkbox `lock instead of removing`?

creesch:
Sounds like the most simple solution yeah.

creesch:
That's even better yeah, that way we can even make the default choice an option.

PR #153, opened 2019-10-06
feat(macros): Add lock comment functionality to modmacros

SpyTec:
Part of #94. Adds lock comment functionality to modmacros

`lockThread()` has been renamed to `lock()`
`unlockThread()` has been renamed to `unlock()`
`lockThread()` now points to `lock()`
`unlockThread()` now points to `unlock()`

## Creating modmacro
image

## Editing modmacro
image

## Triggering macro
image

## Result
image

### DIFF
--- a/extension/data/tbutils.js
+++ b/extension/data/tbutils.js
@@ -2387,7 +2387,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 });
         };
 
-        TBUtils.lockThread = function (id, callback) {
+        TBUtils.lock = function (id, callback) {
             TBUtils.post('/api/lock', {
                 id,
                 uh: TBUtils.modhash,
@@ -2404,7 +2404,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 });
         };
 
-        TBUtils.unlockThread = function (id, callback) {
+        TBUtils.unlock = function (id, callback) {
             TBUtils.post('/api/unlock', {
                 uh: TBUtils.modhash,
                 id,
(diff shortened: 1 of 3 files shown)


extension/data/tbutils.js
```diff
@@ -2387,7 +2387,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 });
         };
 
-        TBUtils.lockThread = function (id, callback) {
+        TBUtils.lock = function (id, callback) {
```
SpyTec: Should have `lockComment` and `lockThread? instead`

Geo1088: Since it's only one method in the API, it makes sense to keep it as is.

creesch:
I am thinking the wording in some areas might need some adjustments. 
In the creation/edit windows possibly change it to `lock macro comment` to make clear it is about that. 
When posting a macro I just noticed that it says `this post will be locked` which is also confusing might  change that to `Parent post will be locked`? @Geo1088 thoughts?

SpyTec:
Or just `lock reply` or `lock removal reply` would work as well. Not sure I can see how `this post will be locked` would be confused with something, what do you think people would see it as if not the post?

creesch:
`lock reply` would work yeah. 


`This post will be locked` took me a moment to parse as the other `This`   items in the screenshot apply to the macro. It isn't a big deal though as it it is what it currently states.

SpyTec:
I wonder, should we also make it possible to lock the comment being removed? At the moment the macro only locks the reply

creesch:
Actually... I just noticed something it does. `lockthread` already locks the parent and we never bothered to check if the parent is a post or comment so it works on both. So I suggest we do the following: 

- On the UI side of things in creation/editing rename it to `lock parent` 
- When applying the macro make it fill in `comment` or `post` in the UI depending on what the parent is.

SpyTec:
Sounds good. I'll fix it up and commit it here

SpyTec:
Fix up

PR #159, opened 2019-10-08
feat(Modbar): Expand new modmail number

### LINKED ISSUE #147: Draw attention to new modmail in modbar if there as an highlighted item present.
If something has been highlighted the assumption is that it is important for the entire team. So drawing some extra attention to it might get mods that don't check modmail as often to check it for those case. I am thinking about either a different color or highlight yellow exclamation mark next to the counter.

SpyTec:
fixes #147.

I'm unsure about this one, I've applied some spans for the various elements when it's expanded, but it could be made better. Thoughts?

This just makes it divided into three:
image

### DIFF
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -30,6 +30,11 @@ function modbar () {
         default: true,
         title: 'Include a button in the modbar to swap between old and new Reddit',
     });
+    self.register_setting('expandedNewModmailCount', {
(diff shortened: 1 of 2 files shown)


creesch:
I am still not sure about this implementation as a solution so I am not going to merge it just yet.

creesch:
Going to close this one as we are likely not doing it like this.

PR #163, opened 2019-10-10
feat(Nuke): Ability to lock instead of remove

### LINKED ISSUE #94: Add comment lock functionality to more toolbox functions

SpyTec:
Fixes #94

Does not have a config entry, instead always defaults to remove. If we want to make this have a default option in settings we should implement a radio config type

image


creesch:
> settings we should implement a radio config type 

Not really, we can simply use the existing selector setting type, [example here](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/e868bea31b61afc5a129f643fa18be41abd6184a/extension/data/modules/historybutton.js#L26-L32)

SpyTec:
I'll add that then and let you when know PR is ready :)

PR #164, opened 2019-10-11
fix(Modbar): don't count highlighted in count

### LINKED ISSUE #162: bug(Modbar): New Modmail count includes highlighted
Right now it includes highlighted, which will be added to the count with in progress, or new and will thus report twice. 

This is one modmail, not two
image

SpyTec:
fixes #162

### DIFF
--- a/extension/data/modules/notifier.js
+++ b/extension/data/modules/notifier.js
@@ -336,7 +336,7 @@ function notifiermod () {
                 setTimeout(() => {
                     TBUtils.apiOauthGET('/api/mod/conversations/unread/count').then(response => {
                         const data = response.data;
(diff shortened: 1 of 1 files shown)


SpyTec:
Or maybe there's more to this issue. Is it possible that someone highlights and archives a modmail and it still exists shows up in count endpoint?

creesch:
> Is it possible that someone highlights and archives a modmail and it still exists shows up in count endpoint?

Possibly, it is probably easier to just leave the count as it is.

SpyTec:
If it is the case, then yes. No way we can figure out whether to remove highlighted or not - would be a different matter if highlighted was counted separately (#159) ;)

creesch:
Checked it, seems to be okay.

PR #170, opened 2019-10-13
fix(RemovalReasons): Hide "Add removal reason" button in modqueue

### LINKED ISSUE #169: "add removal reason" shows on comments where it shouldn't.
The button shouldn't show: 

- In queues. 
- When the user doesn't have removal reasons enabled for comments.

SpyTec:
fixes #169 

Also don't show "Add removal reason" for comments if the user setting is disabled

### DIFF
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -161,21 +161,19 @@ function removalreasons () {
 
         // UI components
         // UI event handling
-        if (TBUtils.pageDetails.pageType !== 'queueListing') {
-            TB.listener.on('post', e => {
-                if (e.detail.data.isRemoved) {
+        TB.listener.on('post', e => {
+            if (e.detail.data.isRemoved && TBUtils.pageDetails.pageType !== 'queueListing') {
+                const $target = $(e.target);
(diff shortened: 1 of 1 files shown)


extension/data/modules/removalreasons.js
```diff
@@ -161,19 +161,21 @@ function removalreasons () {
 
         // UI components
         // UI event handling
-        TB.listener.on('post', e => {
-            if (e.detail.data.isRemoved) {
-                const $target = $(e.target);
-                $target.append(`<span class="tb-bracket-button tb-add-removal-reason" data-id="${e.detail.data.id}" data-subreddit="${e.detail.data.subreddit.name}">Add removal reason</span>`);
-            }
-        });
+        if (TBUtils.pageDetails.pageType !== 'queueListing') {
```
creesch: This is at the wrong place, on redesign if someone starts with a queue page they never will get the notification. It needs to be switched with the line below.  And done separately for the comment listener.

SpyTec: Redesign, right. I'll move it to the listeners

SpyTec:
Fixed feedback in latest commit

extension/data/modules/removalreasons.js
creesch: 👍

Issue #44, opened 2019-04-07
Change webext API calls to use browser instead of chrome


Geo1088:
Firefox implements a promise-based version of Chrome's `chrome.*` extension API, referenced as `browser.*`: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API

It also provides a polyfill to support Chrome: https://github.com/mozilla/webextension-polyfill

Given we're converting lots of our code to promises, I think it would make sense to use a promise-based browser API since it doesn't seem like it's otherwise much different (shouldn't require much logic rewriting, just cleanup).

creesch:
Yeah the problem with that used to be edge, with edge switching to chromium that hopefully shouldn't be an issue anymore.

PR #171, opened 2019-10-14
Background page structuring

### LINKED ISSUE #129: Possibly structure background code a bit more.
I am not sure it is needed (yet), but we have been putting more stuff in background.js and it might be worth it to structure it a bit more and possibly split it up according to functionality. 

Just making this issue so we can discuss it and so I don't forget.

Geo1088:
Fixes #129, partially addresses #44.

TODOs:

- [x] Split up various concerns in the background page to their own files for organization
- [ ] Rework message format from `{action: 'thing', ...options}` to `['thing', {options}]` to free up the `action` key for use in handlers
- [ ] Rename message actions to not use the `tb-` prefix and use a more descriptive convention (needs discussion)
- [x] `chrome.*` -> `browser.*` webextension APIs


extension/data/background/handlers/cache.js
```diff
@@ -0,0 +1,156 @@
+/* global messageHandlers */
```
SpyTec: What's this for?

extension/data/background/handlers/cache.js
```diff
@@ -0,0 +1,156 @@
...
+            browser.tabs.sendMessage(tabs[i].id, {
+                action: 'tb-cache-timeout',
+                payload: cacheType,
+            });
+        }
+    }
+
+    // Done, go for another round.
+    cachedata.timeouts[cacheType] = setTimeout(() => {
+        emptyCacheTimeout(timeoutDuration, cacheType);
```
SpyTec: You've made the function async, you should add an await here

extension/data/background/handlers/cache.js
```diff
@@ -0,0 +1,156 @@
...
+        if (typeof storageShortLength !== 'number') {
+            storageShortLength = parseInt(storageShortLength);
+        }
+    }
+
+    // Compare the current timeout value to the one in storage. Reinit timeout when needed.
+    if (storageShortLength !== cachedata.currentDurations.short || forceRefresh) {
+        console.log('Short timeout', storageShortLength);
+        cachedata.currentDurations.short = storageShortLength;
+        emptyCacheTimeout(storageShortLength, 'short');
```
SpyTec: Same here, function is now `async` so you need this to be `await`

extension/data/background/handlers/cache.js
```diff
@@ -0,0 +1,156 @@
...
+        if (typeof storageLongLength !== 'number') {
+            storageLongLength = parseInt(storageLongLength);
+        }
+    }
+
+    // Compare the current timeout value to the one in storage. Reinit timeout when needed.
+    if (storageLongLength !== cachedata.currentDurations.long || forceRefresh) {
+        console.log('Long timeout', storageLongLength);
+        cachedata.currentDurations.short = storageShortLength;
+        emptyCacheTimeout(storageLongLength, 'long');
```
SpyTec: `await`

extension/data/background/handlers/notifications.js
```diff
@@ -0,0 +1,160 @@
...
+    return notificationID;
+}
+
+// Handle notification creation
+messageHandlers.set('tb-notification', request => {
+    const notificationTimeout = 6000;
+    const sendNotification = request.native ? sendNativeNotification : sendPageNotification;
+    sendNotification(request.details).then(id => {
+        setTimeout(() => {
+            clearNotification(id);
```
SpyTec: Function is now `async` so you need this to be `await`

extension/data/background/handlers/notifications.js
```diff
@@ -0,0 +1,160 @@
...
+
+    // Open up in new tab.
+    const window = await browser.windows.getLastFocused();
+    browser.tabs.create({
+        url: metadata.url,
+        windowId: window.id,
+    });
+
+    // Notification no longer needed, clear it.
+    clearNotification(notificationID);
```
SpyTec: Function is now `async` so you need this to be `await`

extension/data/background/handlers/notifications.js
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Handle click/close events on in-page notifications
+messageHandlers.set('tb-page-notification-click', request => {
+    onClickNotification(request.id);
+});
+
+messageHandlers.set('tb-page-notification-close', request => {
+    clearNotification(request.id);
```
SpyTec: `await`

extension/data/background/handlers/webrequest.js
```diff
@@ -0,0 +1,123 @@
+/* global messageHandlers */
+'use strict';
+
+/**
+ * Retrieves the user's OAuth tokens from cookies.
+ * @param {number?} [tries=1] Number of tries to get the token (recursive)
```
SpyTec: The `[]` denotes optional. Question mark means nothing
```suggestion
 * @param {number} [tries=1] Number of tries to get the token (recursive)
```

.eslintrc.json
```diff
@@ -17,6 +17,7 @@
         "CodeMirror": true,
         "redditapi": true,
         "chrome": true,
+        "browser": true,
```
creesch: odd we never specified browser. Did you run into something that made you add this? (I am also seeing safari defined, probably something we want to remove at some point?)

extension/data/background/background.js
```diff
@@ -1,488 +1,20 @@
...
-    const storageLongLengthKey = 'Toolbox.Utils.longLength';
-    let storageShortLength;
-    let storageLongLength;
-
-    // Get current shortLength value from storage.
-    if (TBsettingsObject.tbsettings[storageShortLengthKey] === undefined) {
-        storageShortLength = 15;
-    } else {
-        storageShortLength = TBsettingsObject.tbsettings[storageShortLengthKey];
+const messageHandlers = new Map();
```
creesch: Is there a specific reason you decided to use a map here? I suspect it mostly has to do with the relatively clean syntax of accessing things through it? 

Mostly curious, I think the usage itself is absolutely fine :)

extension/data/background/handlers/cache.js
creesch: messageHandlers is defined in a different file so without this eslint will complain. He can't add it to the eslint config because it is only global in the background. 

We might want to explore having a different eslint config for background pages now that I think about it...

extension/data/background/handlers/globalmessage.js
```diff
@@ -0,0 +1,29 @@
...
+    // Send to all tabs
+    const tabs = await browser.tabs.query({});
+    for (let i = 0; i < tabs.length; ++i) {
+        if (sender.tab.id !== tabs[i].id && tabs[i].url.includes('reddit.com')) {
+            browser.tabs.sendMessage(tabs[i].id, message);
+        }
+    }
+
+    // HACK: stuff here is defined in the cache file, ideally we shouldn't need
+    //       to call these things across files but there seems to be no way to
```
creesch: Why not simply use regular events? I just did a simple test with `background2.js` and `background.js` where one contained 
```
window.addEventListener('testEvent', () => {
    console.log('test event');
});
```

and the other 

```
const event = new CustomEvent('testEvent');
window.dispatchEvent(event);
```

Which seems to work fine, unless I am misunderstanding the issue here?

extension/data/background/handlers/notifications.js
```diff
@@ -0,0 +1,160 @@
+/* global messageHandlers */
+'use strict';
+// Notification stuff
+
+// We store notification meta data here for later use.
+const notificationData = {};
+
+// TODO: I know we've had this conversation before but I'm 99% sure this isn't
```
creesch: You are still right but we left it in just because with firefox you can't be certain about these things. Better safe than sorry. 

Here is the discord convo we had last time we talked about it. 

```
Geo☆1088 - 06/16/2019
the docs say you can pass any string :V

creesch - 06/16/2019
But to prevent other weird shit, having certainty that is unique helps
the docs are bad for notifications
because they also don't say it is mandatory for firefox, or at the time didn't say thaty

Geo☆1088 - 06/16/2019
the other way i did it also guarantees uniqueness, it was just simpler
and it's not mandatory for firefox either in the testing i did
it works fine just using the string notification-id

creesch - 06/16/2019
I am talking internal uniqueness
With chrome you don't need to pass a parameter at all is what I am saying
seriously, there are some odd differences between the two
Anyway, it might be fine in testing and then we get people to actually use it and it turns out they have ten other extensions using notifications who also using simply a higher number and suddenly you got conflicts
I am not saying I want it to be unique for toolbox, I want it to be unique for the entire browser context

Geo☆1088 - 06/16/2019
okay, yeah, but then we just use tb-${lastID++} :P
i get what you're saying though, don't change more stuff than we have to when working with the notifications

creesch - 06/16/2019
And then we get to fix incognito mode split and firefox still gets confused

Geo☆1088 - 06/16/2019
fair
i don't think the notification ids can confict across extension scopes anyway but i haven't tested that
i guess i wouldn't be surprised if they did

creesch - 06/16/2019
I don't know either but this is firefox we are talking about :wink:

Geo☆1088 - 06/16/2019
lmao
```

extension/data/background/handlers/webrequest.js
```diff
@@ -0,0 +1,123 @@
...
+ */
+function getOAuthTokens (tries = 1) {
+    return new Promise(async (resolve, reject) => {
+        // This function will fetch the cookie and if there is no cookie attempt to create one by visiting modmail.
+        // http://stackoverflow.com/questions/20077487/chrome-extension-message-passing-response-not-sent
+        const rawCookie = await browser.cookies.get({url: 'https://mod.reddit.com', name: 'token'});
+        // If we do get a rawcookie we first want to make sure it is still valid.
+        let expired = false;
+        if (rawCookie) {
+            const cookieExpiration = new Date(rawCookie.expirationDate * 1000).valueOf();
```
creesch: I made some changes here based on your feedback you want to merge in at some point.

extension/data/background/handlers/webrequest.js
```diff
@@ -0,0 +1,123 @@
...
+            resolve({data, textStatus, jqXHR});
+        }, (jqXHR, textStatus, errorThrown) => {
+            jqXHR.allResponseHeaders = makeHeaderObject(jqXHR.getAllResponseHeaders());
+            resolve({jqXHR, textStatus, errorThrown});
+        });
+    });
+}
+
+messageHandlers.set('tb-request', async request => {
+    // TODO: this is a misuse of JSDoc but at least it highlights in VS Code
```
creesch: I know this was already in there but I am not sure what the TODO is actually :D

extension/manifest.json
```diff
@@ -31,8 +31,14 @@
     },
     "background": {
         "scripts": [
+            "data/libs/browser-polyfill-0.5.0.min.js",
```
creesch: We don't need to include this when using firefox, right?  So removing this line from the manifest when building for firefox is probably a good idea.

extension/data/background/handlers/cache.js
creesch: messageHandlers is defined in a different file so without this eslint will complain. He can't add it to the eslint config because it is only global in the background.

We might want to explore having a different eslint config for background pages now that I think about it...

extension/data/background/handlers/cache.js
Geo1088: The await is unneeded since we're not concerned with the resolution of this call. Same is true of other places you left this comment.

extension/data/background/handlers/webrequest.js
Geo1088: This was unchanged from before this PR, will update

.eslintrc.json
Geo1088: Using `browser` without adding `/* global browser */` at the top of each file. We're gonna be doing this conversion for the content scripts eventually too, so it does apply to the whole project.

Can definitely remove the safari definition, l'll do that

creesch: Oh yeah um disregard, I was thinking about browser context not browser the webextension API 😅

extension/data/background/background.js
Geo1088: Yeah, nice syntax and I think map lookups are a bit faster than native object lookups for this use case since you don't have to go through a prototype chain.

extension/data/background/handlers/globalmessage.js
Geo1088: Good call, I'll implement it that way.

extension/data/background/background.js
creesch: Makes sense, might be worth it to implement this instead of objects in other parts of toolbox as well. I think we have a fair amount of places where we use `key`, `value` lists. I am going to guess at least in the places where we currently have the eslint `hasOwnProperty` warnings.

extension/data/background/handlers/notifications.js
Geo1088: yeah I looked up that conversation after I pushed this code, lol. It's still kinda bothering me that we have this random chunk of code that hasn't been touched in ages and uses the crypto api for what could just be a normal incrementing id (or in the case of native notifications, we can even have the browser pick an id for us) but I wouldn't really want to change it without extensive testing. figured i'd mark it down as something we can ideally change in the future when things are more consistent, but probably not in scope for this merge.

extension/data/background/handlers/webrequest.js
Geo1088: Will do

extension/data/background/handlers/webrequest.js
Geo1088: Right, the doclet underneath defines properties on the passed object in a way that I don't think jsdoc handles correctly. I'll look into this.

extension/manifest.json
Geo1088: It's a no-op under firefox, I'd prefer to always include it and let it do its own compatibility testing rather than assume certain browsers support certain calls natively.

creesch: No-op as in not operational? I figured as much and it probably is fine from a technical point of view. I am more concerned with the reviewers at mozilla who sometimes have the tendency to be weird about shit like this. I can already see the review comment saying something like 

> This library is not active in firefox and should not be included. 

Or something stupid like that. 

Also, it is a polyfill for native firefox functionality so I'd expect it to work properly. Like I said I mostly want to avoid dealing with mozilla reviewers being batshit crazy silly again about technicalities.

extension/data/background/handlers/globalmessage.js
Geo1088: I've actually implemented this a slightly different way to maintain the performance benefits of Maps over event listeners.

extension/data/background/handlers/webrequest.js
Geo1088: Also changed `timeNow > cookieExpiration ? true : false` to just `timeNow > cookieExpiration` because `>` always has a boolean result :P

extension/data/background/handlers/webrequest.js
Geo1088: There's no good way to represent this in JSDoc, honestly... Should I just nuke the tag? It would probably be better to document this in TBHelpers or something like that since it only gets called from content scripts.

PR #175, opened 2019-10-15
split tbutils into tbcore, tbhelpers and tbapi

creesch:
I decided to split things up in three new objects instead of keeping tbutils. In order: 

1.  `TBHelpers` included in[`tbhelpers.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/175/files#diff-67e55e5d4c2a111a861c21c89a373e0a) contains helper functions that aren't depending on anything else. Things from array functions to token replacement, etc. Basically our own little lodash.
2. `TBApi` included in [`tbapi.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/175/files#diff-67e55e5d4c2a111a861c21c89a373e0a) contains the majority of the API wrapper functions from `tbutils` excluding those that are explicitly wrapped into a `tbutils` info function
3. `TBCore` included in [`tbcore.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/175/files#diff-a76b1964e0993e8c0194fdc7b17ad978) contains everything else that was included in `tbutils` which basically amounts to statics, other info functions, core functionality and triggers hence the name. 

The approach used in creating this is that I first created 1. and 2. moved all the needed functions to there and replaced their object prefix. I then replaced their use instances in the rest of toolbox. After all that was done I replaced all leftover instances of `TBUtils` with `TBCore`. 

I should note there are some instances of `utils` left like for example the `SETTINGS_NAME` constant under which util settings used to be stored. I don't think we need to migrate that to be honest. 

I did some initial testing and so far everything seems to work fine. With testing I made sure to make use of functionality that touches all three areas. 

TODO: 

- [ ] Update documentation


creesch:
I am still contemplating what to do with our chunking functions. At the moment I am inclined to house regular chunking and dynamic chunking in the helper object and the ratelimit one in the api one.

SpyTec:
> e.

Sounds good to me!

PR #180, opened 2019-10-20
Prevent modbutton popup in flatview from closing flatview

### LINKED ISSUE #168: Closing modbutton popup in flatview closes flatview as well.
Probably an easy fix as it doesn't happen in other overlays.

k3n:
Fixes #168.

### DIFF
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -210,7 +210,7 @@ function modbutton () {
                 $appendTo = $('body');
             }
 
-            let $popup = TB.ui.popup({
+            const $popup = TB.ui.popup({
                 title: `Mod Actions  - /u/${user}`,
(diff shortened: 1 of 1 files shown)


extension/data/modules/modbutton.js
```diff
@@ -428,6 +428,12 @@ function modbutton () {
 
             // reset button name.
             $benbutton.text(self.buttonName);
+
+            // 'cancel' button clicked
+            $popup.on('click', '.close', () => {
+                $popup.remove();
```
creesch: In jQuery `remove()` also cleans up all references so setting it to `null` isn't really needed afaik.

k3n: I removed it, a single occurrence in a low-use area shouldn't make a big difference either way.

PR #182, opened 2019-10-20
Fix new modmail race condition

### LINKED ISSUE #181: Race condition in new modmail
This manifests itself as an error in the console:

```js
Uncaught TypeError: Cannot read property '1' of null
    at Object.TBUtils.getThingInfo(tbutils.js: 1496)
    at addNewMMMacro(macros.js: 144)
    at macros.js: 235
```

Though no actual adverse side-effects are observed, it's best not to have stuff show up in the console.

Repro:

1. nav to new modmail
1. click into a thread
1. _quickly_ click `<` to return to the index
1. observe above error in console*

_\* only appears in Chrome, even though FF also produces the same error condition; I don't use FF much so I'm not sure what's going on, as it seems it should also throw an error._

Seems to be caused by the `idRegex` of `/.*mod.reddit.com/mail/.*?/(.*?)$/i` trying to match the current URL, which under the above repro steps, will simply be `https://mod.reddit.com/mail/notifications` which causes the regex to fail and the ensuing reference to the `[1]` element of the match to throw the error.

k3n:
Fixes #181

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -844,8 +844,10 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 const idRegex = new RegExp('.*mod.reddit.com/mail/.*?/(.*?)$', 'i');
 
                 subreddit = $body.find('.ThreadTitle__community').text();
-                permalink = $threadBase.find('.m-link').length ? `https://mod.reddit.com${$threadBase.find('.m-link').attr('href')}` : `https://mod.reddit.com/mail/perma/${browserUrl.match(idRegex)[1]}`;
-                id = browserUrl.match(idRegex)[1];
+                const idMatch = browserUrl.match(idRegex);
+                // `idMatch` can be null when quickly navigating away (in which case `id` is inconsequential)
+                id = idMatch && idMatch[1] || 'racey';
+                permalink = $threadBase.find('.m-link').length ? `https://mod.reddit.com${$threadBase.find('.m-link').attr('href')}` : `https://mod.reddit.com/mail/perma/${id}`;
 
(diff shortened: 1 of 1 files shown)


extension/data/tbcore.js
```diff
@@ -844,8 +844,10 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 const idRegex = new RegExp('.*mod.reddit.com/mail/.*?/(.*?)$', 'i');
 
                 subreddit = $body.find('.ThreadTitle__community').text();
-                permalink = $threadBase.find('.m-link').length ? `https://mod.reddit.com${$threadBase.find('.m-link').attr('href')}` : `https://mod.reddit.com/mail/perma/${browserUrl.match(idRegex)[1]}`;
-                id = browserUrl.match(idRegex)[1];
+                const idMatch = browserUrl.match(idRegex);
+                // `idMatch` can be null when quickly navigating away (in which case `id` is inconsequential)
+                id = idMatch && idMatch[1] || 'racey';
```
creesch: Does this actually resolve into an id? To me it seems like id well now always be either `true` or `racey`?

k3n: JS uses an novel 'last value' return for `&&` and `||` -- they don't produce bools unless the value of the last fulfilled condition is a bool. I can redo as a ternary to make it more obvious if you'd like, no problem.

creesch: Oh that's pretty interesting. But for readability I do think that a ternary would be better yeah.

Geo1088:
```suggestion
                id = idMatch ? idMatch[1] : 'racey';
```

Issue #104, opened 2019-06-17
Idea: Implementation of tests


SpyTec:
I believe we've always wanted tests, though never had a chance to implement it ourselves. After some discussion on Discord we've decided we want to add unit tests, and e2e tests.

## Test tools
To make testing easier we should use similar setup as RES. 

* For e2e we can utilize [Nightwatch](https://nightwatchjs.org/).
  * creesch has worked with it and believe it's good
  * It has dashboard functionality, how useful it is I don't know. But would give some insight in how it looks when a test fails for example.
  * Could not find a plugin for extension testing, we should look at how RES does it.
* For unit RES uses AVA, though I've only worked with Jest, and Mocha+sinonjs. I'm up for discussion

## How
When it comes to the tests themselves, we have a few approaches I would suggest

* Unit tests should mock out any Reddit responses and work with successes and errors alike. Tests should be written as a black box, data goes in and data or requests comes out.
* E2e tests should utilize both live and mocked responses. As in, we can have a live test to make sure toolbox integrates well with the API, and we can have tests without server to test failures.
* CI that runs tests on every commit. If possible it should only run the e2e tests when something is released to avoid attacking the API too much

Once we have established what we will do we will create a tracking issue for what needs to be setup in terms of enabling tooling, code coverage, and CI. [Codecov.io](https://codecov.io/) can be used for code coverage reporting. The tracking issue should also detail all the components that need tests written for them, preferably in order.

creesch:
# e2e/UI testing

> Could not find a plugin for extension testing, we should look at how RES does it.

Seems to be part of the selenium calls, no plugin needed just provide the browser with a profile that contains the extension. You can see how they did it [here specifically](https://github.com/honestbleeps/Reddit-Enhancement-Suite/blob/master/nightwatch.conf.js). 

Which they run with travis where you can simply see the console output as dashboard. 

For example: https://travis-ci.org/honestbleeps/Reddit-Enhancement-Suite/jobs/546448318




## Sorts of tests I'd like to have. 
Obviously I don't want to go overboard here with the sorts of tests we do here, I'd like to limit it to a few important functionalities that work through a variety of UI triggers and have broken down in the past. Some things that come to mind. 

- Usernotes:
   - Adding a usernote.
     - Viewing added usernote in the button and the window itself. 
     - Viewing same notes in a fresh installation (cleared settings). 
   - Removing a usernote. 

[...]

PR #185, opened 2019-10-21
Unit test framework

SpyTec:
Root `tests/unit` folder for tests. Tests themselves need to import the file and trigger an event to call the function

Part of #104

### DIFF
--- a/tests/unit/modules/support.test.js
+++ b/tests/unit/modules/support.test.js
@@ -3,10 +3,12 @@ require('../../../extension/data/modules/support');
 
 describe('support.js', () => {
     it('calls TB.Module', () => {
+        // Mock module and register_module
         window.TB = {
             Module: jest.fn(),
             register_module: jest.fn(),
         };
+        // What's returned when calling new TB.Module('name')
         window.TB.Module.mockImplementation(() => ({
             shortname: 'support',
             settings: {
@@ -16,8 +18,9 @@ describe('support.js', () => {
                 },
             },
         }));
+        // Dispatch an event to call the function
         const event = new CustomEvent('TBModuleLoaded');
         window.dispatchEvent(event);
-        expect(window.TB.Module).toHaveBeenCalled();
+        expect(window.TB.Module).toHaveBeenCalledWith('Support Module');
     });
 });
--- a/tests/unit/modules/support.test.js
+++ b/tests/unit/modules/support.test.js
@@ -28,8 +28,8 @@ describe('support.js', () => {
             window.dispatchEvent(event);
             window.TB.Module.mock.results[0].value.init();
         });
-        it('calls updates textArea with debug info on r/toolbox link submit', () => {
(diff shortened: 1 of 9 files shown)


tests/unit/modules/support.test.js
```diff
@@ -0,0 +1,58 @@
...
+        expect(window.TB.Module.mock.results[0].value.settings.enabled.default).toBeTruthy();
+    });
+    it('sets name to enabled.hidden to true', () => {
+        const event = new CustomEvent('TBModuleLoaded');
+        window.dispatchEvent(event);
+        expect(window.TB.Module.mock.results[0].value.settings.enabled.hidden).toBeTruthy();
+    });
+    describe('init()', () => {
+        beforeEach(() => {
+            window.TBCore.debugInformation.mockClear;
```
k3n: Is this supposed to be a method call?

tests/unit/modules/support.test.js
```diff
@@ -0,0 +1,58 @@
...
+        });
+        it('calls TBCore.debugInformation()', () => {
+            const event = new CustomEvent('TBModuleLoaded');
+            window.dispatchEvent(event);
+            window.TB.Module.mock.results[0].value.init();
+            expect(window.TBCore.debugInformation).toHaveBeenCalled();
+        });
+        it('calls updates textArea with debug info on r/toolbox link submit', () => {
+            window.history.pushState({}, 'Page Title', '/r/toolbox/submit');
```
k3n: It looks like this isn't getting cleaned up after the test -- `window.history.length === 1` prior to this line, and `2` afterwards... and if you add another test it's still `2`.

tests/unit/modules/support.test.js
creesch: Sonarcloud [agrees](https://sonarcloud.io/project/issues?id=toolbox-team_reddit-moderator-toolbox&open=AW3wZwzm1OSLGlZ3SMXq&pullRequest=185&resolved=false&types=BUG) with @k3n here ;)

creesch:
Had a quick lookover and looks good to me.

Probably a discussion for something outside of this PR but I am wondering if we should start building tests for modules or maybe first for the utility classes and objects?

tests/unit/modules/support.test.js
SpyTec: Yes!

tests/unit/modules/support.test.js
SpyTec: I think we should be using `replaceState` instead to avoid this

SpyTec:
I'm gonna see if we can do this in the env file to avoid having to reset mocks ourselves

```js
beforeEach(() => {
    jest.clearAllMocks();
});
```

> Probably a discussion for something outside of this PR but I am wondering if we should start building tests for modules or maybe first for the utility classes and objects?

Answered in #104

Issue #39, opened 2019-02-05
cleanup: Replace jQuery array utilities ($.each, $.inArray) with ES6 equivalents


Geo1088:
Since we're assuming ES6 now, we can make this easier on ourselves:

- `$.each(arr, (i, val) => {})` becomes `arr.each((val, i) => {})`
- `$.inArray(thing, arr) !== -1` becomes `arr.includes(thing)`
- `$.inArray(thing, arr)` becomes `arr.indexOf(thing)`

creesch:
Okay we need to put in some effort of actually doing this otherwise this isn't going to happen. 

One thing though

> * `$.inArray(thing, arr) !== -1` becomes `arr.includes(thing)`
> 
> * `$.inArray(thing, arr)` becomes `arr.indexOf(thing)`

I am fairly sure most if not all there can be replaced with `arr.includes(thing)` or  `!arr.includes(thing)` unless I am missing something.

Geo1088:
Yeah I'm pretty sure that's the most common case. I think there might be a couple cases where we're re-ordering stuff in arrays where we do use `$.inArray` to get an actual index rather than just whether or not an element exists though.

Either way I expect it won't take long to get this done, I can do a quick run-through of some of this in a bit.

PR #188, opened 2019-10-23
Misc cleanup: domaintagger

k3n:
This started as work on #39 but I ended up addressing a couple other things relating to the style guide and general idiomatic JS. If there's not an appetite for this currently I understand, was just bored since I couldn't find any low-hanging bugs to knock out.


creesch:
Looks good to me, Thanks!

PR #189, opened 2019-10-24
Fix old modmail "loaded in: NaN"

### LINKED ISSUE #138: Mod mail loaded in: NaN seconds
Using Microsoft Edge Dev (yes) and the chromium reddit-moderator-toolbox, on the modmail I get the following message : « mod mail loaded in: NaN seconds ».

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/b3e7442e28f64c1d0f547a4601d1ba12622117b0/extension/data/modules/modmailpro.js#L692

k3n:
Fixes #138.

### DIFF
--- a/extension/data/modules/modmailpro.js
+++ b/extension/data/modules/modmailpro.js
@@ -249,6 +249,7 @@ function modmailpro () {
         $menuList.after($mmpMenu);
         self.endProfile('menu-gen');
 
+        let profileStart = performance.now();
         initialize();
 
         // Processing functions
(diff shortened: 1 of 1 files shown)


extension/data/modules/modmailpro.js
```diff
@@ -249,6 +249,7 @@ function modmailpro () {
         $menuList.after($mmpMenu);
         self.endProfile('menu-gen');
 
+        let profileStart = performance.now();
```
k3n: There's 2 other places this is assigned a value, but they're both within click callbacks, so I'm not sure how this worked previously but this seems to fix it.

creesch: Cool, good enough for me. This is a legacy module anyway which we only keep around for those grumpy mods refusing to use new modmail ;)

PR #190, opened 2019-10-25
Add "use strict" everywhere

SpyTec:
Adds use strict everywhere. Nothing seems to be broken when looking around.

This also changes use strict to error instead of warn


creesch:
Alright, we'll need to bit the bullet at some point anyway. Let's do this.

PR #191, opened 2019-10-27
Some tbmodule tests

k3n:
I've never used Jest, and so decided to plug away at some code I was curious about. Tests are far from comprehensive but are a good start I think.

I welcome any & all feedback, there's some work to do with respect to how & where mocks should be handled, etc. I'll explain some decisions inline.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,10 +1,8 @@
 {
-    "plugins": ["jest"],
     "parserOptions": {
         "ecmaVersion": 2017
     },
     "env": {
-        "jest/globals": true,
         "browser": true,
         "jquery": true,
         "es6": true
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,7 +32,7 @@
     "extends": "eslint:recommended",
     "rules": {
         "strict": [
-            "warn",
+            "error",
             "global"
         ],
         "no-prototype-builtins": "warn",
--- a/package.json
+++ b/package.json
@@ -41,6 +41,10 @@
         ],
         "setupFilesAfterEnv": [
             "./tests/unit/jest.setup.js"
+        ],
+        "coveragePathIgnorePatterns" : [
+            "/node_modules/",
+            "./extension/data/libs/"
         ]
     }
 }
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
             "/node_modules/",
             "./extension/data/libs/"
         ],
+        "collectCoverageFrom" : ["./extension/data/**/*.js"],
         "moduleNameMapper": {
           "^@data(.*)$": "<rootDir>/extension/data$1"
         }
--- a/package.json
+++ b/package.json
@@ -45,6 +45,9 @@
         "coveragePathIgnorePatterns" : [
             "/node_modules/",
             "./extension/data/libs/"
-        ]
+        ],
+        "moduleNameMapper": {
+          "^@data(.*)$": "<rootDir>/extension/data$1"
+        }
     }
 }
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
         "docs": "jsdoc -c jsdoc.json",
         "test": "jest",
         "test:coverage": "npm run test -- --coverage",
-        "codecov": "codecov"
+        "codecov": "codecov",
+        "coveralls": "cat ./coverage/lcov.info | coveralls"
     },
     "homepage": "http://www.reddit.com/r/toolbox",
     "repository": {
@@ -21,6 +22,7 @@
     "devDependencies": {
         "archiver": "^3.1.1",
         "codecov": "^3.6.1",
+        "coveralls": "^3.0.7",
         "docdash": "^1.1.1",
         "eslint": "^5.16.0",
         "eslint-plugin-jest": "^22.19.0",
@@ -42,13 +44,15 @@
         "setupFilesAfterEnv": [
             "./tests/unit/jest.setup.js"
         ],
-        "coveragePathIgnorePatterns" : [
+        "coveragePathIgnorePatterns": [
             "/node_modules/",
             "./extension/data/libs/"
         ],
-        "collectCoverageFrom" : ["./extension/data/**/*.js"],
+        "collectCoverageFrom": [
+            "./extension/data/**/*.js"
+        ],
         "moduleNameMapper": {
-          "^@data(.*)$": "<rootDir>/extension/data$1"
+            "^@data(.*)$": "<rootDir>/extension/data$1"
         }
     }
 }
--- a/tests/unit/jest.setup.js
+++ b/tests/unit/jest.setup.js
@@ -1,33 +1,12 @@
 'use strict';
+
 window.$ = require('../../extension/data/libs/jquery-3.4.1');
+Object.assign(window.$, {
+    log: jest.fn(),
+});
 
 beforeEach(() => {
     jest.clearAllMocks();
     document.body.innerHTML = undefined;
     $('body').off();
 });
-
-// Mock module and register_module
-window.TB = {
-    // What's returned when calling new TB.Module('name')
-    Module: jest.fn().mockImplementation(() => ({
-        shortname: 'support',
-        settings: {
-            enabled: {
-                default: null,
-                hidden: null,
-            },
-        },
-        log: jest.fn(),
-    })),
-    register_module: jest.fn(),
-};
-
-const debugInformationField = jest.fn().mockReturnValue('debugInfoField');
-window.TBCore = {
-    debugInformation: jest.fn().mockReturnValue(debugInformationField),
-};
-
-window.TBHelpers = {
-    template: jest.fn().mockReturnValue('test response'),
-};
--- a/tests/unit/modules/support.test.js
+++ b/tests/unit/modules/support.test.js
@@ -1,7 +1,12 @@
 'use strict';
 require('../../../extension/data/modules/support');
+const {mockWindow} = require('../mocksTBCore');
 
 describe('support.js', () => {
+    beforeAll(() => {
+        Object.assign(window, mockWindow);
+    });
+
     it('calls TB.Module with "Support Module"', () => {
         const event = new CustomEvent('TBModuleLoaded');
         window.dispatchEvent(event);
@@ -27,6 +32,7 @@ describe('support.js', () => {
             const event = new CustomEvent('TBModuleLoaded');
             window.dispatchEvent(event);
             window.TB.Module.mock.results[0].value.init();
+            expect(window.TBCore.debugInformation).toHaveBeenCalledTimes(1);
         });
         it('updates textArea with debug info on r/toolbox link submit', () => {
             window.history.replaceState({}, 'Page Title', '/r/toolbox/submit');
--- a/tests/unit/tbmodule.test.js
+++ b/tests/unit/tbmodule.test.js
@@ -1,0 +1,518 @@
+'use strict';
+
+require('../../extension/data/tbmodule');
+
+describe('tbmodule.js', () => {
+    const setupMockWindow = target => {
+        const profileResults = jest.fn();
+
+        const TBCore = {
+            debugMode: false,
+            devMode: false,
+            isOldReddit: false,
+            // showSettings()...
+            // advancedMode: false,
+            // devModeLock: false,
+            // RandomQuote: '',
+            // toolboxVersion: '',
+            // releaseName: '',
+            // catchEvent: jest.fn(),
+            // clearCache: jest.fn(),
+            // link: jest.fn(),
+            // importSettings: jest.fn(),
+            // exportSettings: jest.fn(),
+            // sendEvent: jest.fn(),
+            // mySubs: [],
+            // and more...
+        };
+
+        const TBui = {
+            // showSettings()...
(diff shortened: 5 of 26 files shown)


tests/__mocks__/tbmodule.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
```
k3n: This file contains mocks for many of the core objects, but file is named `tbmodule` because that's the set of tests I was addressing. This probably could be generalized, but it feels dirty having to create such extensive mocks. Then again the current architecture kind of dictates it.

tests/jest.setup.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+window.$ = require('../extension/data/libs/jquery-3.4.1');
```
k3n: Since this is the core setup file (can only declare 1 in `package.json`), I tried to make it as generic as possible. Most of the stuff that was in here conflicted with the tests I was writing.

tests/unit/modules/support.test.js
```diff
@@ -2,6 +2,39 @@
 require('../../../extension/data/modules/support');
 
 describe('support.js', () => {
+    beforeAll(() => {
+        // Mock module and register_module
+        window.TB = {
```
k3n: Since these mocks were fairly short, I included them inline even though that may not be preferred long-term. I just needed to move them out of the `jest.setup.js` file. @SpyTec

creesch:
I had a quick look over and what I noticed is that the test descriptions you used are fairly abstract where I personally prefer tests to be a bit more in BDDish style where the `it` also describes the expected outcome somewhat. That allows people to quickly scan the tests to see for what purpose they are written without having to look at the asserts themselves and figure it out if that makes sense? 


Unrelated to this specific commit but we need to configure sonarcloud to ignore tests. 

I'll try to find some more time to look at the test.

.eslintrc.json
```diff
@@ -1,10 +1,8 @@
 {
-    "plugins": ["jest"],
     "parserOptions": {
         "ecmaVersion": 2017
     },
     "env": {
-        "jest/globals": true,
```
SpyTec: Why was these removed? They're not in `tests/.eslintrc.json` either

package.json
```diff
@@ -40,7 +40,7 @@
             "jest-webextension-mock"
         ],
         "setupFilesAfterEnv": [
-            "./tests/unit/jest.setup.js"
+            "<rootDir>/tests/jest.setup.js"
```
SpyTec: `<rootDir>` only works if you set it in a string, see https://jestjs.io/docs/en/configuration#rootdir-string

tests/unit/jest.setup.js
```diff
@@ -1,33 +0,0 @@
...
-};
-
-const debugInformationField = jest.fn().mockReturnValue('debugInfoField');
-window.TBCore = {
-    debugInformation: jest.fn().mockReturnValue(debugInformationField),
-};
-
-window.TBHelpers = {
-    template: jest.fn().mockReturnValue('test response'),
-};
```
SpyTec: Please keep it in /unit, for e2e we will use something else (see the dedicated branch)

tests/unit/modules/support.test.js
SpyTec: Not sure I follow, why do they need to be moved out of `jest.setup.js`?

tests/unit/tbmodule.test.js
```diff
@@ -0,0 +1,175 @@
...
+describe('tbmodule.js', () => {
+    describe('tbmodule', () => {
+        const dispatchTBCoreLoaded = () => window.dispatchEvent(new CustomEvent('TBCoreLoaded'));
+
+        beforeAll(() => {
+            Object.assign(window, mockWindow);
+        });
+
+        afterEach(() => {
+            jest.clearAllMocks();
```
SpyTec: We will want to clear this globally every time due to stuff like jQuery, no point having it in test files

tests/unit/tbmodule.test.js
```diff
@@ -0,0 +1,175 @@
...
+            expect(tbModEvent.type).toStrictEqual(eventName);
+        });
+
+        describe('`window.TB` API', () => {
+            beforeEach(() => {
+                dispatchTBCoreLoaded();
+            });
+
+            it('properties', () => {
+                expect.assertions(5);
```
SpyTec: Didn't know about this, learning something new every day

tests/__mocks__/tbmodule.js
SpyTec: You mean to say this is called every time `tbmodule` is imported in a test? Unless it mocks something that isn't used globally it should be in a global file

tests/unit/modules/support.test.js
```diff
@@ -2,6 +2,39 @@
...
+
+        window.TBHelpers = {
+            template: jest.fn().mockReturnValue('test response'),
+        };
+    });
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+        document.body.innerHTML = undefined;
+        $('body').off();
```
SpyTec: Any reason we shouldn't call this globally?

tests/__mocks__/tbmodule.js
k3n: I couldn't get the automagical inclusion of mocks working; the Jest guides want you to put them along side in a dir named `__mocks__`, but I'm not a fan of that, would rather they live in the `tests` dir, so I dumped it into here for the time being. As it currently stands, they're just manually imported by `tests/unit/tbmodule.test.js`.

tests/unit/modules/support.test.js
k3n: Quick example: you're mocking `window.TB` here.

However, that object is actually created here in `tbmodule.js`:

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbmodule.js#L3

And I these tests I wrote are to guarantee that `tbmodule.js` creates `window.TB` (along with its properties & methods):

https://github.com/k3n/reddit-moderator-toolbox/blob/test_tbmodule/tests/unit/tbmodule.test.js#L35

So, there's a bug or refactor in `tbmodule.js` which causes `window.TB` to not get created, we can't catch that if we include your mock in the global mocks because you're mocking `window.TB`.

Yes, the `tbmodule.js` tests I wrote would catch that `window.TB` isn't proper (it's missing properties & methods), but that's only because of the the `window.TB` mock, and if someone in the future comes along and improves the `window.TB` mock to include mocks for all the properties & methods, then my tests would be completely redundant.

I hope all that makes sense...?

.eslintrc.json
k3n: Is there a reason to have them in the global eslint? I wouldn't think we'd want that...

For your 2nd point: the `"extends": ["plugin:jest/recommended"]` I added to `tests/.eslintrc.json` covers the Jest globals. And I'm not married to `["plugin:jest/recommended"]` (or `["plugin:jest/all"]` or anything, but I did find many of the requirements useful (especially with `all`).

package.json
k3n: Leftover from trying to make paths more manageable (I get annoyed with `require('../../../../../some/path')`). It works for running tests but I can change it back.

tests/unit/jest.setup.js
k3n: Ah, wasn't aware of the alternate method for e2e. Will move back.

tests/unit/modules/support.test.js
k3n: To better learn how Jest works I wanted to have a clean slate for my tests, but now that I have everything in place I should be able to move most/all of this back to `jest.setup.js`.

tests/unit/modules/support.test.js
k3n: Thinking about this more -- we have, at a minimum, 2 different kinds of JS files in the repo:

1. TB framework code
2. TB modules code

The modules code is obviously everything under `/modules`, where each file creates an instance of a `TB.Module` (and possibly does other stuff). The framework code is most everything else not in `/modules`.

I'm calling out this distinction because the 2 types have different testing needs, e.g. the modules might need stuff like `TB.Module` mocked, whereas the framework code might now. I do think it'd be bad to replicate mocks across each module test, but I also don't think it should be injected globally on account of reasons I gave in my previous comment.

Would it be better to just have master repo of mocks, which we can compose into our desired mock object? For instance, [this code](https://github.com/k3n/reddit-moderator-toolbox/blob/test_tbmodule/tests/__mocks__/tbmodule.js#L70-L78) would instead live inside my test, while the full library of mocked objects would remain as an importable code?

Then, if we need to overrride any specific mock, and can simply not import it (or augment it's imported instance).

tests/__mocks__/tbmodule.js
k3n: Another part of my reasoning is my long-standing practice (I think it's common...) of having your `/tests` dir mirror your app dir. If I were to put `__mocks__` inside `/tests/unit` then that pattern would be lost. Naming [and placing] things is hard! :)

SpyTec: > I couldn't get the automagical inclusion of mocks working; the Jest guides want you to put them alongside app code in a dir named `__mocks__`, but I'm not a fan of that, would rather they live in the `tests` dir, so I dumped it into here for the time being. As it currently stands, they're just manually imported by `tests/unit/tbmodule.test.js`.

@creesch, whats your thoughts on putting `__mocks__/` in `src/`? And @k3n, did you verify that it does automagically import the mock if it's in `src/__mocks__`?

> Another part of my reasoning is my long-standing practice (I think it's common...) of having your `/tests` dir mirror your app dir. If I were to put `__mocks__` inside `/tests/unit` then that pattern would be lost. Naming [and placing] things is hard! :)

Well, whether mirroring app dir happens in `tests/` or `tests/unit` doesn't really matter

tests/unit/modules/support.test.js
SpyTec: Having just a have a moduleMock file that mocks core so we can use it in modules would be my go to. We'd have to import this every time we create a module test file

For core, I'd suggest we don't use any mock files but mock in the test file itself. Mostly as we can't mock that easily.

tests/__mocks__/tbmodule.js
SpyTec: Reading up on it more, I don't think we can use mocks automagically anyhow. It would mock exports, we have none in our code. Putting it in src won't make a difference. For that purpose, we shouldn't use `__mocks__` folder as it won't be accurate to what people expect from that folder

package.json
SpyTec: How would this change imports within tests?

tests/__mocks__/tbmodule.js
creesch: Uhm let's see. Since we are also going to do e2e tests we decided to do `tests/unit/` and `tests/e2/`, so mirroring the app structure in `tests/` isn't going to work. We decided on this approach to cluster the tests but still separate the sort of tests, I honestly don't know what standard practice is here, it just made sense to me. 

For simplicity let's say that we do mirror in `tests/unit/` then I personally do not have an issue with `__mocks__` also living there as they apply directly to the unit test and the directory name makes it very clear what the purpose is. Also if jest recommends that structure then that is also an argument for me to simply follow it.

package.json
k3n: I was hoping that setting `roots: ['<rootDir>/data/extension', '<rootDir>/tests']`` would make it so that we could just do `tbmodule('tbmodule')` in the test and not worry about getting the right amount of dots, but I couldn't get it to work (didn't spend a bunch of time on it).

Btw I think the `<rootDir>` is fine unless you saw a problem from it? The docs you reference just mean that you can re-use it elsewhere it sounds like, though I admit I could be reading that wrong.

k3n:
New day, new commit :)

tests/jest.setup.js
```diff
@@ -0,0 +1,12 @@
...
+Object.assign(window.$, {
+    log: jest.fn(),
+});
+
+beforeEach(() => {
+    jest.clearAllMocks();
+    document.body.innerHTML = undefined;
+    $('body').off();
+});
```
SpyTec: Lets move this to `/unit`, needs to be changed in package.json too

tests/unit/tbmodule.test.js
```diff
@@ -0,0 +1,279 @@
...
+                    });
+                });
+                it('the `settingsList` property is an array with a single string element "enabled"', () => {
+                    expect(module.settingsList).toStrictEqual(['enabled']);
+                });
+            });
+        });
+
+        describe('instance methods', () => {
+            describe('the `register_setting()` method', () => {
```
SpyTec: It's a bit too verbose IMO. We can go with this:
```suggestion
            describe('register_setting()', () => {
```

Only thing that matters really is being verbose in the `it('adds foo')` stuff. You agree @creesch?

tests/jest.setup.js
k3n: I actually did move it (though I failed to update package.json).

SpyTec: I don't see it moved, it's still under `/tests/jest.setup.js` and not `/tests/unit/jest.setup.js`

k3n: Clearly I need to hone my telepathic coding skills...

PR #197, opened 2019-11-20
Convert more stuff to use TBLog

Geo1088:
Replaces `console.*` calls in most of the extension with calls to `TBLog` methods. `tbmigrate.js` still needs to be handled, but I don't know if the profiling function in there right now needs to be redone or not.


creesch:
Looks good to me, I assume this is ready to merge?

SpyTec:
LGTM

PR #213, opened 2020-02-09
Convert remaining `chrome.*` calls to `browser.*`

### LINKED ISSUE #44: Change webext API calls to use browser instead of chrome

Geo1088:
Fixes #44 🎉

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -1463,7 +1463,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
         };
 
         // Listen to background page communication and act based on that.
-        chrome.runtime.onMessage.addListener(message => {
+        browser.runtime.onMessage.addListener(message => {
             switch (message.action) {
             case 'clearCache': {
                 TBCore.clearCache(true);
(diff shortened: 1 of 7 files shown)


extension/data/tbcore.js
```diff
@@ -174,7 +174,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
             TBCore.mySubsData = cacheDetails.moderatedSubsData;
         }
 
-        const manifest = chrome.runtime.getManifest();
+        const manifest = browser.runtime.getManifest();
```
creesch: Does this still work as expected? I'd hope it is still synchronous here but if not we need to fetch it elsewhere.

Geo1088: Yep, I checked the docs for this and similar calls, everything that returns a synchronous result under `chrome.*` is functionally identical under `browser.*`.

PR #214, opened 2020-02-10
Add ability to select by score in queues

Blank-Cheque:
Added ability to select queue items by score greater than or less than a certain number to `queuetools.js`.

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -560,7 +560,7 @@ function queuetools () {
                     break;
                 case 'pointsgt': {
                     const min = parseInt(prompt('points greater than:', ''));
-                    selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] > min;
+                    selector = (_, el) => $(el).find('.score.unvoted').attr('title') > min;
                     break;
                 }
                 case 'pointslt': {
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -558,14 +558,16 @@ function queuetools () {
                 case 'flair':
                     selector = ':has(.linkflairlabel)';
                     break;
-                case 'pointsgt':
-                    min = parseInt(prompt('points greater than:', ''));
+                case 'pointsgt': {
+                    const min = parseInt(prompt('points greater than:', ''));
                     selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] > min;
                     break;
-                case 'pointslt':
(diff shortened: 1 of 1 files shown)


extension/data/modules/queuetools.js
```diff
@@ -556,6 +558,14 @@ function queuetools () {
                 case 'flair':
                     selector = ':has(.linkflairlabel)';
                     break;
+                case 'pointsgt':
+                    min = parseInt(prompt('points greater than:', ''));
+                    selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] > min;
+                    break;
```
Geo1088: Is `min` actually declared anywhere? This should probably be rewritten inside a block with an explicit declaration if this is the only place that variable is being used. For example:

```js
case 'pointsgt': {
	const min = parseInt(prompt('points greater than:', ''));
	selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] > min;
	break;
}
```

Geo1088: Resolved, convo [here](https://discordapp.com/channels/535490452066009090/535490452066009093/676249302217130016)

extension/data/modules/queuetools.js
```diff
@@ -556,6 +558,16 @@ function queuetools () {
                 case 'flair':
                     selector = ':has(.linkflairlabel)';
                     break;
+                case 'pointsgt': {
+                    const min = parseInt(prompt('points greater than:', ''));
+                    selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] > min;
```
Geo1088: Turns out the `innerHTML` here isn't guaranteed to contain an actual number, things can get shortened to e.g. "196k" (check /top for an example). However, the `title` attribute does always have a number. I also simplified the selector a bit with jQuery.

```suggestion
                    selector = (_, el) => $(el).find('.score.unvoted').attr('title') > min;
```

extension/data/modules/queuetools.js
```diff
@@ -556,6 +558,16 @@ function queuetools () {
...
+                    break;
+                }
+                case 'pointslt': {
+                    const max = parseInt(prompt('points less than:', ''));
+                    selector = t => things[t].querySelector('.score.unvoted').innerHTML.split(' ')[0] < max;
```
Geo1088:
```suggestion
                    selector = (_, el) => $(el).find('.score.unvoted').attr('title') < max;
```

Geo1088:
Looks good, all working on my end and no lint errors! Thanks for the contribution.

PR #215, opened 2020-02-10
Add ability to sort subs in queue multiple times

Blank-Cheque:
The previous version explicitly only allowed multi-subreddit queues to be sorted by number of items in each subreddit's queue once. This has been changed for the sake of convenience.

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -787,7 +787,16 @@ function queuetools () {
                 }
 
                 self.log('sorting queue sidebar');
-                $('.tb-sort-subs').remove(); // don't allow sorting twice.
+
+                var oldBoxes = document.querySelectorAll('.tb-subreddit-item-count');
+                var oldBoxesLen = oldBoxes.length
+                for (var x = 0; x < oldBoxesLen; x++) {
+                    oldBoxes[0].remove();
+                };
+
+                const sortButton = document.querySelector('.tb-sort-subs');
+                sortButton.innerHTML = 'sorting...';
+                sortButton.style.cssText = 'padding-left: 17px; padding-right: 16px;';
 
                 const now = TBHelpers.getTime(),
                     // delay = 0,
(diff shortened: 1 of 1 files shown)


extension/data/modules/queuetools.js
```diff
@@ -787,7 +787,16 @@ function queuetools () {
                 }
 
                 self.log('sorting queue sidebar');
-                $('.tb-sort-subs').remove(); // don't allow sorting twice.
+
+                var oldBoxes = document.querySelectorAll('.tb-subreddit-item-count');
+                var oldBoxesLen = oldBoxes.length
+                for (var x = 0; x < oldBoxesLen; x++) {
+                    oldBoxes[0].remove();
```
k3n: For this `for()` loop to work, this would need to be `oldBoxes[x].remove();`, but with jQuery the whole block could be reduced to just `$('.tb-subreddit-item-count').remove();`.

extension/data/modules/queuetools.js
```diff
@@ -787,7 +787,16 @@ function queuetools () {
                 }
 
                 self.log('sorting queue sidebar');
-                $('.tb-sort-subs').remove(); // don't allow sorting twice.
+
+                var oldBoxes = document.querySelectorAll('.tb-subreddit-item-count');
+                var oldBoxesLen = oldBoxes.length
```
SpyTec: `oldBoxes` can be a `const` here. It's never changed, then remove `oldBoxesLen` based off of @k3n's comment

extension/data/modules/queuetools.js
Blank-Cheque: As the array's length continually changes due to elements being removed from it, removing `oldBoxes[x]` will not result in every element being removed; it will result in half of them being removed (as seen in this screenshot), and then a TypeError being thrown
image
Instead we remove the first element repeatedly until they're all gone.

Blank-Cheque: Thank you for your suggestion on reducing that to one line, I've updated the PR in accordance.

k3n: You're correct about the array index, I had forgotten that QSA actually returns a `NodeList` which is a live collection, as opposed to a static array of elements. Thanks for keeping me honest :)

PR #223, opened 2020-02-12
Usernotes manager rework

### LINKED ISSUE #221: Usernotes manager performs badly
Just fix all of it tbh, labeling as a bug because it basically makes the usernotes page unusable for subs that are near the limit.

Geo1088:
Fixes #221. Will require heavy review.

Adds a `TBui.pager` helper for creating paged content. Reworks the usernote manager's rendering routine to use that helper so that not all notes are rendered at once. This prevents the browser from hanging for long periods of time on subreddits with tons of notes.

Still a lot to do:

- [x] Improve pagination interface with overwhelming amounts of pages
- [ ] ~~Allow user to set how many users per page to display~~ scrapped, not really necessary
- [x] Make search etc. work without searching the DOM
- [x] Add options for pruning by note age


Geo1088:
Screenshot of the pagination interface as of last commit: 
image

creesch:
Do we want to move this to 5.5 as a goal @Geo1088 ? I am currently looking into what is needed to release 5.4

Geo1088:
Would really rather not, this is a fix that people have told me a bunch of times they're waiting on. I can go over this branch soon and break any tasks I haven't c completed yet into issues for a future release, but the core of this issue should definitely get fixed for 5.4.

creesch:
Fair enough, any insights in time availability from your side of things? If it is possible to do this in a stable manner in a relative short time frame I'd be okay with waiting on that. On the other hand if it is going to cost some time due to your availability we can always shortly release 5.5 after 5.4 just for this fix. 

Which reminds me that I really need to start looking into automated publishing and figuring out what is needed for the various stores to have multiple people registered to publish as well. Because then (if we lay of big refactoring for a bit ;)) we can easily do single feature releases (rolling releases effectively).

Geo1088:
No reason I couldn't get this done today or tomorrow. I've been dedicating most of my time lately to new tooling needed for /r/anime, but if we're trying to release soon now I can get this done quickly.

creesch:
> but if we're trying to release soon now 

Well we have been sitting on a bunch of stuff so I figured we'd at least move that to the users soon :) 

> I can get this done quickly.

Awesome!

Geo1088:
Should be stable now.

The last thing I was working on was some changes to usernote pruning. I've only implemented the simplest of the three strategies I wanted - pruning by the age of the notes themselves. This is easy because there's no need to make API requests to check account activity/availability. The other two I plan to get to soon, but I don't think I will before next release. They're still in the UI as a reminder to myself to do them, but disabled for now.

Geo1088:
Deepscan and Travis are both getting hung up on some async syntax and I'm not sure why...? [Travis fails on an eslint parsing error](https://travis-ci.org/github/toolbox-team/reddit-moderator-toolbox/jobs/715994023) but `npm run eslint` works fine on my machine. My environment matches except I'm on node 12, but I don't think that should matter...?

creesch:
@Geo1088 are you going to have time to look into the things I found. I'd really like to get a version out. As a sidenote, I wouldn't mind doing a special release only for this feature either.

Geo1088:
> Did a quick test and now when opening the manager all usernames show up as `undefined` so something is quite right yet.

I still can't reproduce this. Are you sure you were testing the latest commits?

I think I'll open a separate PR for case-sensitivity stuff after this gets merged.

creesch:
> > Did a quick test and now when opening the manager all usernames show up as `undefined` so something is quite right yet.
> 
> I still can't reproduce this. Are you sure you were testing the latest commits?
> 
> I think I'll open a separate PR for case-sensitivity stuff after this gets merged.

Yeah I am sure I was running on the latest commits. Trying again now it might just be when opening usernote manager again directly after it has ran. When I clear my toolbox cache it shows properly again. 


Repo steps are:  

1. Open usernote manager.
2. Prune notes.
3. After it is done immediately open it. 


Just now I also had one instance where there was just one note left for some mysterious reason where there should roughly be 300. Haven't been able to reproduce that one though...

Geo1088:
> Repo steps are:  

> 1. Open usernote manager.
> 2. Prune notes.
> 3. After it is done immediately open it. 

Thanks, reproduced.

Geo1088:
Fixed now. The pruning operation basically rewrites the usernotes object from the ground up, and I was reconstructing it without the `name` property on each user. That didn't impact the data saved to the wiki, since that value gets ignored when deflating the object, but the missing key was still written to cache, resulting in bad display until cache was cleared.

Geo1088:
Merged master to pull in `okOnly` fixes and implemented pruning by account activity. This should be set for final review now.

Issue #219, opened 2020-02-11
Removal Reason settings per subreddit


SpyTec:
There's a few things we currently have user-side that could also exist in subreddit settings.

- [ ] Sticky removal message
- [ ] Lock removal message
- [ ] Remove as comment/PM/Modmail
- [x] Enable/Disable removal reason for posts (#224)
- [x] Enable/Disable removal reason for comments (#224)

creesch:
I thought we did discuss this already at some point, I guess we didn't make an issue for it at the time? 

Additional question to figure out: 

- If a subreddit has set a specific configuration should the individual mod still be able to overwrite it? (for example, still send through modmail). 
- If not, should we provide it as a explicit setting "`allow mods to override subreddit`"?

creesch:
Also:

> Allow/Disallow removal reason for posts
> Allow/Disallow removal reason for comments

Shouldn't these be enable/disable? Should we consider requiring at least one to be active?

SpyTec:
> I thought we did discuss this already at some point, I guess we didn't make an issue for it at the time?
We did, couldn't find an issue about it

> Additional question to figure out:
> 
>     * If a subreddit has set a specific configuration should the individual mod still be able to overwrite it? (for example, still send through modmail).
>     * If not, should we provide it as a explicit setting "`allow mods to override subreddit`"?
I'd say yes, we should only pre-select the values. Though if mods really want to disallow certain actions we could add another setting for disabling actions. Such as remove as PM or modmail.



> Also:
> 
> > Allow/Disallow removal reason for posts
> > Allow/Disallow removal reason for comments
> 
> Shouldn't these be enable/disable? Should we consider requiring at least one to be active? Because disabling both effectively turns of removal reasons which I can already see resulting in us getting odd support requests.
Right, yes. That's what I meant :D

Geo1088:
>  Enable/Disable removal reason for posts
>  Enable/Disable removal reason for comments

I'd much rather see these as settings per-removal reason than per-sub, and not allow the user to override them. The "no reason" button will always exist and there's no good reason for mods to blanket ignore removal reasons if the team took the time to set them up. Some removal reasons apply to comments but not posts, and this system would provide better flexibility.

>  Sticky removal message
>  Lock removal message
>  Remove as comment/PM/Modmail

[...]

PR #224, opened 2020-02-12
feat(RR): Per post/comment RRs

SpyTec:
Part of #219. 

Implements ability to have removal reasons per post and/or comments. (With bonus feature that if both are empty, it is not visible when removing)

By default, if the RR doesn't have `removePosts` and `removeComments` set we will show the RR for both types. If it exists it's a boolean to show for posts and/or comments.

If user has `commentReasons` setting disabled, we won't show RR for comments unless explicitly set to `true` in settings


creesch:
I forgot, is this complete?

SpyTec:
Yes, sir

PR #230, opened 2020-02-19
add modmail preview functionality fixes #229

### LINKED ISSUE #229: message preview in modmail
RES doesn't work in modmail and we do include the needed library to render markdown in reddit style, so it should be fairly doable to include this. 

Easiest is probably to do it in github style with a preview tab, though doing it live would also be neat.

creesch:
What it says on the tin. Adds a live preview area on the text area that updates through a new debounce utility function (currently set at 500ms).

### DIFF
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -102,7 +102,7 @@ function newmodmailpro () {
                 $body.one('click', '.ThreadViewerReplyForm__replyButton, .NewThread__submitButton', () => {
                     $previewArea.remove();
                 });
-            }, 500));
+            }, 100));
         }
 
         if (searchhelp) {
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -45,6 +45,12 @@ function newmodmailpro () {
         title: 'Automatically switch "reply as" selection away from "Reply as myself" to "Reply as subreddit".',
     });
 
+    self.register_setting('showModmailPreview', {
+        type: 'boolean',
+        default: true,
+        title: 'Show a preview of modmail messages while typing.',
+    });
+
     const $body = $('body');
 
     function switchAwayFromReplyAsSelf () {
@@ -58,12 +64,14 @@ function newmodmailpro () {
     if (TBCore.isNewModmail) {
         // Add a class to body
         $body.addClass('tb-new-modmail');
+        const parser = SnuOwnd.getParser(SnuOwnd.getRedditRenderer());
 
         // ready some variables.
         const modMailNightmode = self.setting('modmailnightmode'),
               lastReplyTypeCheck = self.setting('lastreplytypecheck'),
               searchhelp = self.setting('searchhelp'),
-              noReplyAsSelf = self.setting('noReplyAsSelf');
+              noReplyAsSelf = self.setting('noReplyAsSelf'),
+              showModmailPreview = self.setting('showModmailPreview');
 
         if (noReplyAsSelf) {
             switchAwayFromReplyAsSelf();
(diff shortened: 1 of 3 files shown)


extension/data/styles/newmodmailpro.css
```diff
@@ -19,6 +19,24 @@
     position: absolute;
 }
 
+/* preview
+*/
+.NewThread  #tb-modmail-preview {
```
SpyTec:
```suggestion
.NewThread #tb-modmail-preview {
```

extension/data/modules/newmodmailpro.js
```diff
@@ -58,12 +64,14 @@ function newmodmailpro () {
...
         $body.addClass('tb-new-modmail');
+        const parser = SnuOwnd.getParser(SnuOwnd.getRedditRenderer());
 
         // ready some variables.
         const modMailNightmode = self.setting('modmailnightmode'),
               lastReplyTypeCheck = self.setting('lastreplytypecheck'),
               searchhelp = self.setting('searchhelp'),
-              noReplyAsSelf = self.setting('noReplyAsSelf');
+              noReplyAsSelf = self.setting('noReplyAsSelf'),
+              showModmailPreview = self.setting('showModmailPreview');
```
SpyTec: Can we start declaring each variable on its own instead of one big const declaration? We'd only change one line then and not two every time

creesch: That's not the convention we are using, so we can discuss it but that is outside the scope of this PR ;)

extension/data/tbhelpers.js
```diff
@@ -2,6 +2,25 @@
 /** @namespace  TBHelpers */
 (function (TBHelpers) {
     const logger = TBLog('TBHelpers');
+
+    /**
+     * Debounces a given function based on a given timeout.
+     * @function debounce
+     * @memberof TBHelpers
+     * @param {function} func input function
+     * @param {integer} debounceTime the amount of time used to wait in ms.
```
Geo1088: `integer` isn't a jsdoc type, this should just be `number`

extension/data/modules/newmodmailpro.js
```diff
@@ -58,12 +64,14 @@ function newmodmailpro () {
     if (TBCore.isNewModmail) {
         // Add a class to body
         $body.addClass('tb-new-modmail');
+        const parser = SnuOwnd.getParser(SnuOwnd.getRedditRenderer());
```
Geo1088: Now that I actually look at it, `snuownd` is very outdated and won't support stuff like spoilers or any other markdown changes that have happened since 2015. We should look into implementing the [`snudown-js`](https://github.com/erikdesjardins/snudown-js) module that RES uses instead, since erikdesjardins still seems to maintain that.

extension/data/styles/newmodmailpro.css
```diff
@@ -19,6 +19,24 @@
...
+  
+#tb-modmail-preview {
+    background: #ffffffd1;
+    padding: 10px;
+    box-shadow: 0 1px 4px #efefed;
+}
+
+#tb-modmail-preview h3.tb-preview-heading {
+    border-bottom: solid 1px #c8c7c7;
+    margin-bottom: 15px;
```
Geo1088: Haven't actually tested this yet to see what it looks like, but I don't think Reddit adds a border like this normally, so I don't think we should either.

extension/data/modules/newmodmailpro.js
creesch: Fair enough, though I doubt spoilers work in modmail (we should test that). I am going to leave that outside of this PR if you don't mind and make a maintenance issue for that.

extension/data/tbhelpers.js
SpyTec:
```suggestion
     * @param {number} debounceTime the amount of time used to wait in ms.
```

extension/data/styles/newmodmailpro.css
creesch: That is just the border under a heading we place to indicate it is a preview. See [this video of how it works](https://www.creesch.com/dump/vid/vid_5e4e388f5c205.mp4), without the border under the text the heading seems to belong to the text below it. So it is a toolbox styling thing ;)

Geo1088: Cool, looks good then :+1: thanks for the video

extension/data/tbhelpers.js
creesch: Hrm... could have sworn we use integer elsewhere. I'll change it though.

extension/data/modules/newmodmailpro.js
Geo1088: Even if they don't work from the modmail interface, I'm pretty sure they would still show up correctly for users. Not sure what the use case of sending spoiler tags in modmail might be but you never know.

Either way though I'm fine with making this another issue as long as we address it before the next release.

PR #234, opened 2020-02-25
feat(QueueTools): Better sticky on scroll

SpyTec:
Does not yet work as expected. Moves the sticky content due to `margin` and `left`. And queue items move up slightly

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -326,6 +326,7 @@ function queuetools () {
 
             // Add checkboxes, tabs, menu, etc
             $('#siteTable').before(`
+    <div class="modtools-duplicate" style="display: none; visibility: hidden;"></div>
     <div class="menuarea modtools" style="padding: 5px 0;margin: 5px 0;top: 0px">
         <input style="margin:5px;float:left" title="Select all/none" type="checkbox" id="select-all" title="select all/none"/>
         <span>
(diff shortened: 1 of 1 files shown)


creesch:
Just had a look at it and it shouldn't stick the thing at the top of the screen, there should be a bit of a margin there. It probably should be the same as the space on the left of the thing.

extension/data/modules/queuetools.js
```diff
@@ -430,24 +431,26 @@ function queuetools () {
...
+                frame = requestAnimationFrame(() => {
                     $modtoolsMenu.css({
-                        top: 'inherit',
+                        left: position === 'fixed' ? offsetSticky : 0,
+                        right: position === 'fixed' ? offsetSticky : 0,
+                        top: position === 'fixed' ? offsetSticky : 0,
+                        position,
                     });
-                }
+                    $('#siteTable').css({marginTop: position === 'fixed' ? $modtoolsMenu.outerHeight(true) + 6 : 0});
```
Geo1088: Haven't looked at the full changeset yet but...

I know you and creesch were talking about this in Discord, how there should be space between the top of the bar and the top of the viewport. Previously this was handled automatically, because the CSS for that element sets a `margin-top` and this code only messed with the `top` CSS property. That makes things a lot cleaner than manipularing `margin-top` because then part of the CSS that's in the related stylesheet gets overwritten in an unpredictable way. I'd recommend working this so `top` is the CSS property you're changing.

SpyTec: You mean changing `#siteTable` to use `top` instead of `margin-top`?

Geo1088: I think I actually just completely misunderstood this code, so disregard my first comment, sorry.

What is this `+6` accounting for?

SpyTec: I haven't figured out where that comes from. Adding it makes things work fine. Without it and it jumps 6px when you start scrolling down and it becomes fixed.

Hence why this is still a draft :P

Geo1088: Ah okay, interesting. I'll see if I can make any sense of it later and post my findings in the Discord.

creesch:
So I saw still some talking about this, is it ready or does it need work?

SpyTec:
It is ready

PR #235, opened 2020-02-25
fix(UserNotes): Wrap on resize

### LINKED ISSUE #216: User note type display gets messy with lots of types or small windows
Example

SpyTec:
Fixes #216 
This now uses flexbox and goes to new line when box is too small

### DIFF
--- a/extension/data/styles/usernotes.css
+++ b/extension/data/styles/usernotes.css
@@ -97,7 +97,9 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
     display: flex;
     flex-wrap: wrap;
 }
-
+.utagger-type-list > div {
+  margin: 2px 2px 0 0;
+}
 .mod-toolbox-rd .utagger-popup .utagger-type input {
     display: none;
 }
--- a/extension/data/styles/usernotes.css
+++ b/extension/data/styles/usernotes.css
@@ -104,30 +104,30 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
     display: none;
 }
 
-.mod-toolbox-rd .utagger-popup .utagger-type > * {
+.mod-toolbox-rd .utagger-popup .utagger-type > div {
     padding: 3px;
     border: 1px solid rgb(114, 134, 154);
-    transition: backgroun-color 0.15s ease-out;
+    transition: background-color 100ms ease-out;
 }
 
-.mod-toolbox-rd .utagger-popup td:not(:last-child) .utagger-type span {
+.mod-toolbox-rd .utagger-popup td:not(:last-child) .utagger-type div {
     border-right: none;
 }
 
-.mod-toolbox-rd .utagger-popup .utagger-type > *:hover {
+.mod-toolbox-rd .utagger-popup .utagger-type > div:hover {
     background-color: #e2eefa;
     cursor: pointer;
 }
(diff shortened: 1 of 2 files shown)


Geo1088:
Since this isn't a table anymore and borders between items no longer collapse into each other, could we put like a 2px `margin-left` and `margin-bottom` on each note type to make it look a bit cleaner?

Before: 
image

After: 
image

extension/data/styles/usernotes.css
```diff
@@ -93,11 +93,16 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
...
+
 .mod-toolbox-rd .utagger-popup .utagger-type input {
     display: none;
 }
 
-.mod-toolbox-rd .utagger-popup .utagger-type span {
+.mod-toolbox-rd .utagger-popup .utagger-type > * {
     padding: 3px;
     border: 1px solid rgb(114, 134, 154);
     transition: backgroun-color 0.15s ease-out;
```
creesch: Seems like this was already broken but might as well fix it as the transition as it is now will not work.

extension/data/styles/usernotes.css
```diff
@@ -93,11 +93,16 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
     padding-top: 4px;
 }
 
+.mod-toolbox-rd .utagger-popup .utagger-types .utagger-type-list{
+    display: flex;
+    flex-wrap: wrap;
+}
+
```
creesch:
```suggestion
.utagger-type-list > div {
  margin: 2px 2px 0 0;
}
```

Adds the margin @Geo1088 suggested

extension/data/styles/usernotes.css
```diff
@@ -93,11 +93,16 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
...
 .mod-toolbox-rd .utagger-popup .utagger-type input {
     display: none;
 }
 
-.mod-toolbox-rd .utagger-popup .utagger-type span {
+.mod-toolbox-rd .utagger-popup .utagger-type > * {
```
creesch: Why are you using a wildcard here? Seems to me that `div` should be sufficient?

extension/data/styles/usernotes.css
```diff
@@ -107,20 +112,20 @@ body.mod-toolbox-rd.mod-toolbox-rd-new-modmail .tb-recents .tb-bracket-button {
     border-right: none;
 }
 
-.mod-toolbox-rd .utagger-popup .utagger-type span:hover {
+.mod-toolbox-rd .utagger-popup .utagger-type > *:hover {
     background-color: #e2eefa;
     cursor: pointer;
 }
 .mod-toolbox-rd .utagger-popup .utagger-type input:checked ~ span {
```
creesch: You forgot to change the `span` to a `div`  (or wildcard) here.  Also on line 132 but I can't comment on that from here.

extension/data/styles/usernotes.css
SpyTec: Makes it not dependent on what DOM element we're using so we can change it in the future without having to worry about breaking CSS. It removes the connectivity between CSS and DOM, though I could've just added a new class

creesch: I honestly rather have it be more explicit, chances are low we will change it and css wildcards do have somewhat of a performance impact.

PR #238, opened 2020-02-27
Better centering of window/dialog titles

Geo1088:
Solving a nitpick issue where dialog titles weren't centered with respect to the entire dialog, but with respect to the part of the header that wasn't already taken up by the close button. Also refactors away from hardcoding padding for the settings window title to be centered nicely.

Method applied for this PR comes from loosely following https://stackoverflow.com/a/32546033/1070107

### DIFF
--- a/extension/data/modules/personalnotes.js
+++ b/extension/data/modules/personalnotes.js
@@ -60,8 +60,6 @@ function personalnotes () {
                 ],
                 cssClass: 'personal-notes-popup',
             }).appendTo('body');
-
-            $body.find('.personal-notes-popup').drag('.personal-notes-popup .tb-popup-title');
         }
 
         // Let's load a note!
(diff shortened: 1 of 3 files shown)


SpyTec:
Getting a scroll on subreddit settings even though it's not needed. Seems to be the background tint causing it

image

Geo1088:
Seems intentional judging by the large `margin-bottom` windows get, either way not caused by this PR: 
image

creesch:
Not entirely sure where why the margin there is so large. It is just there to give some spacing in the instances where scroll is actually is needed.The overflow scroll itself is intentional in the sense that overlay windows can be bigger than the viewport.

creesch:
Actually, I didn't realize how small the window there was in height. Seems like it should be around 40px (which is the top value) combined with the scrollbar's 30px so 70px in total.

Geo1088:
The spacing on overlays is actually kinda confusing to me, I'm not sure what element/style actually provides the top margin. Either way though, that can be addressed in another issue.

Geo1088:
oh, this breaks the draggable area for dialog headers, let me fix that before merging

PR #243, opened 2020-03-02
Misc. code cleanup stuff (#243) (rip isOdd)

Geo1088:
Adds documentation to a lot of helper methods, removes useless logic and unused functions, and other stuff like that. No particular priority, I'll be adding to this branch as I go through other code quality issues.


extension/data/modules/modbutton.js
```diff
@@ -551,17 +550,6 @@ function modbutton () {
                 }
             }
 
-            function createBanReason (message) {
```
creesch: huh, what did this ever do?

Geo1088: I have no idea, lol. In its current state it was just a fancy way of returning the same string, so I removed it. Not sure if there was originally other functionality attached that was removed, or if someone had plans to make fancier ban reasons or something like that.

PR #246, opened 2020-03-03
jsAPI flow for sidebar fixes #38

### LINKED ISSUE #38: Explore the possibility of integrating new modmail in the jsapi flow
Currently new modmail is the odd one out that basically does a bit of a hybrid thing between old reddit and new reddit. We probably want to integrate it more into the new reddit jsAPI event flow so relevant buttons are easier to implement in new modmail as well. 

Prompted today with me wanting to use the toolbox profile feature in modmail and finding out that we don't have any way to trigger it there. Which is very silly of course.

creesch:
Very early work in progress. Sidebar flow works as intended but a bunch of work still needs to be done: 

- [x] Adding conversation authors to jsAPI as well so we can remove most really specific new modmail logic from modules. 
- [x] Conversion of button modules to use the new flow and remove the old newmm flow.
- [x] Clean up tbNewthing and possibly remove it as we no longer need it  in this capacity. 
- [ ] Bunch of styling.

### DIFF
--- a/extension/data/modules/historybutton.js
+++ b/extension/data/modules/historybutton.js
@@ -59,20 +59,6 @@ function historybutton () {
         });
     };
 
-    self.attachToModmail = function () {
-        const $body = $('body');
-        $body.find('.Thread__message').not('.tb-history').each(function () {
-            const $this = $(this);
-            if ($this.find('.tb-attr').length === 0) {
-                $this.addClass('tb-history').find('.Message__divider').eq(0).after('<span class="tb-attr"></span>');
-            }
-            const $tbAttrs = $this.find('.tb-attr');
-            const author = $this.find('.Message__header').find('.Message__author').text().substring(2);
-            const UserButtonHTMLnewMM = `<span class="tb-history-button" >&nbsp;<a href="javascript:;" class="user-history-button tb-bracket-button" data-author="${author}" title="view & analyze user\'s submission and comment history">H</a></span>`;
-            $tbAttrs.append(UserButtonHTMLnewMM);
-        });
-    };
-
     self.runJsAPI = function () {
         self.log('run');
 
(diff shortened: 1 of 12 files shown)


extension/data/modules/historybutton.js
```diff
@@ -67,15 +67,10 @@ function historybutton () {
                 $this.addClass('tb-history').find('.Message__divider').eq(0).after('<span class="tb-attr"></span>');
             }
             const $tbAttrs = $this.find('.tb-attr');
-            const UserButtonHTMLnewMM = '<span class="tb-history-button" >&nbsp;<a href="javascript:;" class="user-history-button tb-bracket-button" title="view & analyze user\'s submission and comment history">H</a></span>';
+            const author = $this.find('.Message__header').find('.Message__author').text().substring(2);
```
k3n: Curious why not `find('.Message__header .Message__author')`?

As I understand the code as it is currently, it will query the DOM for each `'.Message__header'` found (not sure if this will be a lot or a little), whereas using a single `find()` will only query once.

creesch:
This is mostly done, except for a css issue with the profile overlay that I haven't figured out yet. 

image

creesch:
I can fix it in modmail by removing the 100% height value for `.tb-window-tabs` but I get the feeling [that the value is in there for a reason](https://github.com/toolbox-team/reddit-moderator-toolbox/blame/master/extension/data/styles/tbui.css#L109). @Geo1088, could use some insight here.

extension/data/modules/oldreddit.js
```diff
@@ -24,6 +24,12 @@ function oldReddit () {
         }
     }
 
+    /**
+     * Handles `thing` items as they become visible in the viewport.
+     * @function
+     * @param {IntersectionObserverEntry} entries
```
Geo1088: Doc nit: this is an array
```suggestion
     * @param {IntersectionObserverEntry[]} entries
```

extension/data/styles/tbui.css
```diff
@@ -106,7 +106,7 @@
 .mod-toolbox-rd .tb-window-tabs {
     display: inline-block;
     width: 135px;
-    height: 100%;
+    height: auto;
```
Geo1088: `height: auto` is the default value, can't this line just be deleted?

creesch: uh, probably :D

extension/data/modules/modbutton.js
```diff
@@ -140,11 +134,7 @@ function modbutton () {
     };
 
     self.init = function () {
-        if (TBCore.isNewModmail) {
-            self.buttonName = 'Mod Button';
-        } else {
-            self.buttonName = 'M';
-        }
+        self.buttonName = 'M';
```
Geo1088: Since `buttonName` is always the same now, I don't think it needs to be a variable here. The "reset" on line 408 should be unnecessary now, this assignment can be removed, and the value can be inlined on line 77.

creesch: Cleaned up the entire thing.

k3n:
Lines changed: `+146 −299` 💯

Issue #248, opened 2020-03-06
Handle ratelimiting consistently


eritbh:
The only real ratelimiting code toolbox actually has is in [TBCore.forEachChunkedRateLimit](), which from what I can tell basically says "here's a function that'll make x API calls, run it y times as fast as you can but don't let the `Ratelimit-Remaining` header dip below 20 and if it does stop everything for six minutes". It was added [back in 2015](https://github.com/toolbox-team/reddit-moderator-toolbox/commit/92298a628feb2aca4bf76c9435a7d1f87e28e99a#diff-c9da5bc7882bb894de2a63a2aaede53aR1027-R1098) and hasn't significantly changed since.

This system provides only very weak protection, relies on tons of assumptions, is written in a very messy and hard-to-understand way, provides no protection for API calls that are made outside this function, and [may just not work as intended at all](https://discordapp.com/channels/535490452066009090/535490452066009093/685253706178101312) when a large amount of requests are required (e.g. usernote pruning by activity).

Anyway enough ranting about that. Ratelimiting should be handled in the background page, by the request handler, by actually reading the ratelimit headers for every request and queueing requests to be sent if more stack up than can be handled in the current ratelimit bucket. Care should be taken to ensure that `ratelimit-reset` headers are interpreted properly so we don't end up waiting longer for ratelimits than we have to.

This system will protect *all* requests from ratelimit-related failure and will obviate `TBCore.forEachChunkedRateLimit`. It can then be removed and replaced by simple `for` loops, and we can make sure that appropriate loading indicators are displayed for all tasks that may take a long time to complete due to frequent limiting.

creesch:
Not going to deny it is messy, though I am not sure where you are getting the 6 minutes from as it should just take the time reddit gives for cooldown. However that very well could be 6 minutes .


Other than that, I am all for a system of handling ratelimiting consistently. Is it maybe an idea to do this as part of #45?

Geo1088:
> though I am not sure where you are getting the 6 minutes from

I'm not sure where it's coming from either, lol. That was an observation I made when testing it yesterday - I posted the logs in discord, `ratelimitReset` was reporting 8 or 9 seconds but the timer told me it was gonna be multiple minutes before anything resumed. The logic behind determining the number of seconds left is very confusing to me and I haven't been able to identify what's actually happening. At this point though, I'm not gonna bother trying to understand the current code, I'll just plan to rip it out and re-implement rate limiting entirely.

[...]

PR #250, opened 2020-03-06
Convert $.ajax() to fetch()

### LINKED ISSUE #45: Convert $.ajax calls to use fetch
Related to #44 

[`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is supported on all relevant browser versions and has a much nicer API than `$.ajax` does. 

Because `$.ajax` passes three parameters to its callbacks, it's more difficult to have it play well with promises, which can only represent a single value (we get around this now by promising an object with the callback values as properties, but this isn't ideal). `fetch`, on the other hand, promises a single [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object.

Additionally, `Response` is (relatively) easily serializable to/from JSON - e.g. we can get the response from the background page, convert it to JSON, and then let `TBUtils.sendRequest` convert the JSON it received from the background page back into a `Response` object with its [constructor](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response). For example:

```js
// background script
browser.runtime.onMessage.addListener(async request => {
    const response = await fetch('...');
    const headers = {};
    for (const [k, v] of response.headers) {
        headers[k] = v;
    }
    return [response.text(), {
        status: response.status,
        statusText: response.statusText,
        headers,
    }];
});
// content script
const response = await browser.runtime.sendMessage({...}).then(responseData => new Response(...responseData));
doSomethingWith(response.json());
```

Geo1088:
Yet another WIP PR from me, fixes #45.

With this change, the API for dealing with requests from modules will change a bit:

- `TBApi.sendRequest()` will resolve to a `Response` regardless of response error code, and will reject an `Error` on network error or request timeout/abortion. This mirrors `fetch()` semantics. Error objects will be always be instances of `Error` rather than e.g. `TypeError` or `NetworkError` because deserializing error objects from JSON is hard without using `eval()`.

- Other request methods will resolve to a `Response` only if the response status is OK. If a non-OK response is received, it will reject with an `Error` object, with the raw response attached as a `response` property. Other network errors that indicate no response was received will be rejected as normal `Error` objects with no `response` property, the same as with `TBApi.sendRequest`.

For example:
```js
TBApi.apiOAuthGET(...).then(async response => {
	// Action was successful!
}).catch(error => {
	if (error.response) {
		// We can check for error codes returned by Reddit here
		self.error('request failed with status', error.response.status);
	} else {
		// The request probably never even got to Reddit
		self.error('non-Reddit error on request:', error.message);
	}
});
```

Additionally, the `makeRequest` function will be made available in the background page. Its API is identical to `TBApi.sendRequest`.

---

A summary of structural changes is in [this comment](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250#issuecomment-597419446).

Known bugs:

- ~~Pulling HTML from JSON is broken somehow, resulting in stuff like the context popups not displaying things correctly. Not sure why this is.~~ Fixed with 6c90bab


extension/data/background/handlers/webrequest.js
```diff
@@ -21,104 +22,120 @@ async function getOAuthTokens (tries = 1) {
...
-    return new Promise(resolve => {
-        $.ajax(options).then((data, textStatus, jqXHR) => {
-            jqXHR.allResponseHeaders = makeHeaderObject(jqXHR.getAllResponseHeaders());
-            resolve({data, textStatus, jqXHR});
-        }, (jqXHR, textStatus, errorThrown) => {
-            jqXHR.allResponseHeaders = makeHeaderObject(jqXHR.getAllResponseHeaders());
-            resolve({jqXHR, textStatus, errorThrown});
-        });
-    });
+function makeRequest (url, options) {
```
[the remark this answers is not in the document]
Geo1088: Commented above, will be used for ratelimit stuff later so might as well keep it in for now. Less work for a future PR.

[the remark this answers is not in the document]
creesch: Gotcha 👍

extension/data/tbapi.js
```diff
@@ -11,26 +11,31 @@
...
+    TBApi.sendRequest = async ({method, endpoint, query, body, oauth}) => {
+        const response = await browser.runtime.sendMessage({
+            action: 'tb-request',
+            method,
+            endpoint,
+            query,
+            body,
+            oauth,
+        });
+        if (response.error) {
```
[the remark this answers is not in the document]
Geo1088: Yes, this implementation is very WIP and nothing will work in its current state.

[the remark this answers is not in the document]
creesch: Well you asked me to go over it :P Other than that I don't spot anything wrong with it at the moment.

Geo1088:
At long last, ready for review! This is another big diff, so here's a summary to make reviewing easier.

- [`extension/data/background/handlers/webrequest.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250/files#diff-0c26dbb2bf6a8eece45cab254fac2dc8)

  - `getOAuthTokens` was simplified to be more readable.
  - `serializeResponse` was created to serialize `Response` objects to JSON so they can be passed to the content script.
  - `makeHeaderObject` was removed, as it's been replaced by part of `serializeResponse`.
  - `makeRequest` was modified to just be a wrapper around `fetch`. This function will contain more stuff later  when we start tackling ratelimiting in #248.
  - The message handler now takes `body` rather than `data`. This prevents confusion between data in the URL parameters and data in the request body.
  - The message handler now messages back either a `Response` serialized to a JSON array, or a JSON object with an `error` key.
  - Various details of the actual requests were changed to match Fetch APIs.

- [`extension/data/tbapi.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250/files#diff-67e55e5d4c2a111a861c21c89a373e0a)
  - `sendRequest` now takes `body` and `query` options rather than a `data` option for the same reason as above.
  - `sendRequest` has a new `okOnly` option. When set to true, the method will throw an error if a non-2xx response error code is encountered. The error will have a `response` property in this case. Otherwise, all responses will be returned regardless of error code, and the only errors thrown will be request interruptions/timeouts/abortions.
  - `apiOAuthRequest` was removed, as it was useless and only used to define `apiOAuthPOST` and `apiOAuthGET`.
  - `getJSON` and `post` were reimplemented with the above changes to lower-level request methods. They still return JSON objects, but now throw `Error` objects rather than the strange object stuff they threw before.
  - `apiOAuthGET` and `apiOAuthPOST` were also reimplemented. They no longer return strange custom objects, instead returning a `Response`. They also throw `Error`s rather than custom objects.

- [`extension/data/modules/historybutton.js](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250/files#diff-170550628f2608d5d6e5d7d99040a3bd)
  - Corrected error handling. When `.catch` precedes `.then` and the `.catch` handler doesn't re-throw, the `.then` is triggered even on error. To compensate for this, we return early if there is no data.

- [`extension/data/modules/modmatrix.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250/files#diff-12edf01b814a98421bbdc86bf2053eb0)
- [`extension/data/modules/notifier.js`](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/250/files#diff-42045f70544a35292ee6ab4fe4fe8025)
  - Updated response and error handling to work with the new expected values (`Response` objects and `Error` objects).

creesch:
Something somewhat related. Am I supposed/allowed to use `makeRequest()` if I want to do requests from a different handler in the background?

Geo1088:
Hm, I didn't really consider that use case. I think I'm gonna change around how things are structured a bit to make that easier to deal with.

Geo1088:
> Am I supposed/allowed to use makeRequest() if I want to do requests from a different handler in the background?

Latest commit, bdd68a2, addresses this. `makeRequest` now handles all options passed in the background page, which means yes, you can just use `makeRequest()` from anywhere else in the background page and pass it the same options you would pass to `TBApi.sendRequest()` in content scripts.

On the background page, `makeRequest` handles all options and makes the request, while the actual message listener adds response serialization for sending the response to the content script. In the content script, `TBApi.sendRequest()` is now solely responsible for sending the request options and deserializing the response.

creesch:
I assume this is still ready for review? Planning to look at this in the next couple of weeks so we can get started with the next release.

Also tagged @SpyTec as this is one of those things we probably want to have an extra set of eyes on.

extension/data/background/handlers/webrequest.js
```diff
@@ -21,104 +22,143 @@ async function getOAuthTokens (tries = 1) {
         cookieInfo.firstPartyDomain = 'reddit.com';
         rawCookie = await browser.cookies.get(cookieInfo);
     }
-    // If we do get a rawcookie we first want to make sure it is still valid.
-    let expired = false;
+
+    // Make sure the cookie is still valid
+    let validCookie = true;
```
SpyTec: Default to `false` and skip the else-statement

extension/data/background/handlers/webrequest.js
```diff
@@ -21,104 +22,143 @@ async function getOAuthTokens (tries = 1) {
...
+    // If we don't have a valid cookie, we need to generate a new one. If the user
+    // is logged in, we can do this by sending a request to the modmail page, and
+    // it'll send back a new cookie that the browser will write for us. We'll then
+    // be able to read the new cookie and all is well, without having to deal with
+    // the Reddit OAuth flow ourselves.
+    if (tries < 3) {
+        await makeRequest('https://mod.reddit.com/mail/all');
+        return getOAuthTokens(tries + 1);
+    } else {
+        // If we tried that 3 times and still no dice, the user probably isn't logged
```
SpyTec: Can't we simply check response code for `401` instead of fetching modmail 3 times?

extension/data/background/handlers/webrequest.js
```diff
@@ -21,104 +22,143 @@ async function getOAuthTokens (tries = 1) {
...
+
+    // Otherwise return the raw response
+    return response;
+}
+
+// Makes a request and sends a reply with response and error properties
+messageHandlers.set('tb-request', requestOptions => makeRequest(requestOptions)
+    .then(async response => {
+        response = await serializeResponse(response);
+        return {response};
```
SpyTec: Do we need await here? And there's no need to re-assign response, `serializeResponse` can just be returned

extension/data/modules/historybutton.js
```diff
@@ -393,6 +393,9 @@ function historybutton () {
             $error.html('unable to load userdata</br>shadowbanned?');
             TB.ui.longLoadNonPersistent(false);
         }).then(d => {
```
SpyTec: Unrelated to PR, is `d` for `data`? Shouldn't have to guess this

extension/data/modules/modmatrix.js
```diff
@@ -434,9 +434,9 @@ function modmatrix () {
                 self.processData(data, callback);
                 self.dataCache[cacheKey] = data;
             })
-                .catch(({jqXHR, textStatus, errorThrown}) => {
-                    self.log(`Mod log request ${requestData.count}to ${requestData.count + requestData.limit} failed (${jqXHR.status}), ${textStatus}: ${errorThrown}`);
-                    if (jqXHR.status === 504) {
+                .catch(error => {
+                    self.log(`Mod log request ${requestData.count}to ${requestData.count + requestData.limit} failed:`, error);
```
SpyTec:
```suggestion
                    self.log(`Mod log request ${requestData.count} to ${requestData.count + requestData.limit} failed:`, error);
```

extension/data/tbapi.js
```diff
@@ -11,88 +11,105 @@
      * Sends a generic HTTP request through the background page.
      * @function
      * @param {object} options The options for the AJAX request
-     * @param {string} options.method The HTTP method to use for the request
+     * @param {string} [options.method] The HTTP method to use for the request
      * @param {string} options.endpoint The endpoint to request
```
SpyTec: Switch `method` and `endpoint` around to have all optional params grouped

extension/data/tbapi.js
```diff
@@ -11,88 +11,105 @@
...
-        method,
-        endpoint,
-        data,
-        oauth,
-    }).then(response => {
-        if (response.errorThrown !== undefined) {
-            throw response;
+    TBApi.sendRequest = async ({method, endpoint, query, body, oauth, okOnly}) => {
+        // Make the request
+        const messageReply = await browser.runtime.sendMessage({
```
SpyTec: Both old and new code never catches in the even it returns an error. [MDN says](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage):

> If an error occurs while connecting to the extension, the promise will be rejected with an error message

Should we not catch this?

extension/data/tbcore.js
```diff
@@ -1812,61 +1812,43 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
...
-                    logger.log('504 Timeout retrying request');
-                    getModSubs(after, subs => callback(modSubs.concat(subs)));
-                } else {
-                    modSubs = [];
-                    return callback(modSubs);
-                }
+        TBApi.getJSON('/subreddits/mine/moderator.json', {
+            after,
+            limit: 100,
+        }).then(data => {
```
SpyTec: Shouldn't this be `response`? Accessing `data.data` is confusing

extension/data/background/handlers/webrequest.js
creesch: No we do this because reddit can be unstable at times and the code we can get in return isn't always reliable.

SpyTec: Typical :P

extension/data/background/handlers/webrequest.js
Geo1088: Await is required because it's being returned as an object with a property. Goal of the reassignment was to increase readability somewhat, but it's a weird structure that's hard to represent clearly because of terminology overlap. Equivalent alternative would be `return {response: await serializeResponse(response)}`, take your pick.

extension/data/modules/historybutton.js
Geo1088: Probably, but won't pretend to know for sure, lol

extension/data/tbapi.js
Geo1088: No, the consumer of sendRequest is meant to handle that case. No need to explicitly re-throw.

extension/data/background/handlers/webrequest.js
SpyTec: I definitely prefer the `{response: await serializeResponse(response)}`. Re-assigning the param is bad practice

extension/data/tbcore.js
Geo1088: `response.data` is potentially confusing because `getJSON` doesn't return a `Response`, but the JSON body. Maybe `json.data` would be clearer?

creesch:
I am getting some errors, but I think those are the result of something already resolved in master related to promisfying some stuff. Can you merge it in so we can double check? 

image

Geo1088:
Done @creesch :+1:

Geo1088:
Context popup issues were caused by bad querystring handling, fixed with latest commit. Can't identify any other bugs with this branch, let's get some final 👀 on this and it should be good to merge!

creesch:
Just did a testdrive hitting a bunch of API related functionality and it works all fine now. I'll await the changes you mentioned that spytec requested and after those (and a tiny bit of testing) I am sure we can merge this.

PR #259, opened 2020-03-12
Queue cache mechanism fixes #187

### LINKED ISSUE #187: queue cache mechanism
Needed for our own overlays to figure out if things are filtered and for RR as it currently removes everything as from redesign it is not clear if things are removed or filtered when clicking "add removal reason".

Cache freshness will be determined by a `lastRefresh` timestamp which will be checked against the thing post or edited timestamp we are looking for.

creesch:
This implements a background page mechanism where the mod queue contents can be requested which will then either be retrieved through the api or given from the caching object. 

It is still a WIP as there are a few things that need to be done: 

- [ ] Convert the jquery getJSON call to fetch once @Geo1088 is done with #250. 
- [ ] Figure out what to do with items that are no longer in the queue but still in the cache. The easiest approach is probably to not only check against the thing name but also let the cache expire after a minute or so. Which is fine as I mostly wanted to prevent a ton of api requests when doing flatview or a profile. I am thinking it probably can be set even shorter in the 10/20 seconds range. 
- [ ] Document stuff a bit more. 
- [ ] Double check the tbUI logic

### DIFF
--- a/extension/data/background/handlers/modqueue.js
+++ b/extension/data/background/handlers/modqueue.js
@@ -1,0 +1,67 @@
+'use strict';
+
+// Object containing the queue cached data per subreddit
+const queueCacheObject = {};
+
+/**
+ * Returns if a given thing is contained in the cache for the given subreddit.
+ * @param thingName reddit `thing` name.
+ * @param subreddit subreddit
+ * @returns {boolean}
+ */
+function thingFound (thingName, subreddit) {
+    let thingFound = false;
+    if (Object.prototype.hasOwnProperty.call(queueCacheObject, subreddit) && queueCacheObject[subreddit].things.includes(thingName)) {
+        thingFound = true;
+    }
+    return thingFound;
+}
+
+messageHandlers.set('tb-modqueue', request => {
+    const {subreddit, thingName, thingTimestamp} = request;
+    return new Promise(resolve => {
+        // Check if we need to fetch data.
+        let lastRefresh = 0;
+        let refreshActive = false;
+        if (Object.prototype.hasOwnProperty.call(queueCacheObject, subreddit)) {
+            lastRefresh = queueCacheObject[subreddit].lastRefresh;
+            refreshActive = queueCacheObject[subreddit].refreshActive;
+        }
+
+        // The thing timestamp is bigger than the last refresh or cache isn't fresh anymore.
+        if (thingTimestamp * 1000 > lastRefresh || Date.now() - lastRefresh > 1000 * 60) {
+            // To reduce api calls we don't do a new one if one is already running for this subreddit.
+            // Instead we wait for the finished event and then reference the cache object.
+            if (refreshActive) {
+                window.addEventListener(`freshCache-${subreddit}`, () => {
+                    resolve(thingFound(thingName, subreddit));
+                }, {
+                    once: true,
+                });
+            } else {
+                if (Object.prototype.hasOwnProperty.call(queueCacheObject, subreddit)) {
+                    queueCacheObject[subreddit].refreshActive = true;
+                } else {
+                    queueCacheObject[subreddit] = {
+                        refreshActive: true,
+                        lastRefresh: 0,
+                        things: [],
+                    };
+                }
+
+                $.getJSON(`https://old.reddit.com/r/${subreddit}/about/modqueue.json?limit=100`).done(updatedQueue => {
+                    const nowRefresh = Date.now();
+                    queueCacheObject[subreddit] = {
+                        lastRefresh: nowRefresh,
+                        things: updatedQueue.data.children.map(thing => thing.data.name),
+                        refreshActive: false,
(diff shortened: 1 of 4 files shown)


creesch:
Right, I should get back on this. I really want to get this into toolbox as people are getting confused in context popup and such.

creesch:
Okay ready enough for review. I did not build in explicit error handling in the sense of retries. It will check for errors and if there are any will resolve with whatever there currently is in queue.

extension/data/background/handlers/modqueue.js
```diff
@@ -0,0 +1,81 @@
...
+const queueCacheObject = {};
+
+/**
+ * Returns if a given thing is contained in the cache for the given subreddit.
+ * @param thingName reddit `thing` name.
+ * @param subreddit subreddit
+ * @returns {boolean}
+ */
+function thingFound (thingName, subreddit) {
+    let thingFound = false;
```
Geo1088, line 13: Name shadowing, might change this variable to `found` so it doesn't match the function name. Did eslint not catch this?

extension/data/background/handlers/modqueue.js
```diff
@@ -0,0 +1,81 @@
...
+    if (Object.prototype.hasOwnProperty.call(queueCacheObject, subreddit) && queueCacheObject[subreddit].things.includes(thingName)) {
```
Geo1088, line 14: I'd consider storing the cache data in a `Map` instead of a plain object to avoid `hasOwnProperty` shenanigans, it's a good structure to use for stuff like this (and background page message handlers already use it anyway).

extension/data/background/handlers/modqueue.js
```diff
@@ -0,0 +1,81 @@
...
+        thingFound = true;
+    }
+    return thingFound;
+}
+
+messageHandlers.set('tb-modqueue', request => {
+    const {subreddit, thingName, thingTimestamp} = request;
+    return new Promise(resolve => {
```
Geo1088, line 22: Could avoid an indentation level by rewriting the function definition as
```js
messageHandlers.set('tb-modqueue', request => new Promise(resolve => {
	const {subreddit, thingName, thingTimestamp} = request;
	// ...
}));
```

extension/data/background/handlers/modqueue.js
```diff
@@ -0,0 +1,81 @@
...
+        // Check if we need to fetch data.
+        let lastRefresh = 0;
+        let refreshActive = false;
+        if (Object.prototype.hasOwnProperty.call(queueCacheObject, subreddit)) {
+            lastRefresh = queueCacheObject[subreddit].lastRefresh;
+            refreshActive = queueCacheObject[subreddit].refreshActive;
+        }
+
+        // The thing timestamp is bigger than the last refresh or cache isn't fresh anymore.
+        if (thingTimestamp * 1000 > lastRefresh || Date.now() - lastRefresh > 1000 * 30) {
```
Geo1088, line 32: might set `30` as a constant (`MODQUEUE_CACHE_TTL` or something to that effect) at the top of the function or file for clarity

extension/data/tbui.js
```diff
@@ -1255,6 +1255,21 @@
             } else {
                 $(`<a class="tb-submission-button tb-submission-button-nsfw" data-fullname="${submissionName}" href="javascript:void(0)">nsfw</a>`).appendTo($submissionButtonList);
             }
+
+            if (submissionStatus === 'removed' || submissionStatus === 'spammed') {
+                browser.runtime.sendMessage({
```
Geo1088, line 1260: I'd probably write a wrapping `TBApi` function rather than using `browser.runtime.sendMessage` directly, though if this functionality isn't going to be used from module code later then it's probably not as important.

extension/data/background/handlers/modqueue.js
creesch: Good catch it did not.

extension/data/tbui.js
creesch: I am not sure what the benefit would be as the params would be the same and it would effectively resolve the same as well.

Geo1088: I prefer to leave the coupling between the background page and the content script all in one place, so if for whatever reason we need to change how the parameters are passed to the background page in the future, we only need to update it in one place rather than changing parameter names in every module it's used in. Not likely to have a huge impact, definitely not important if this won't be used outside `TBui`, just something to consider.

creesch: Fair enough, something for a different issue then as we use it directly in a bunch of places as I never found it needed personally speaking. Makes sense though to put it in one place I suppose.

extension/data/background/handlers/modqueue.js
Geo1088: We probably have the `no-shadow` rule configured incorrectly, I'll look into that in another PR.

PR #270, opened 2020-03-21
Fix settings backup button

### LINKED ISSUE #269: clicking backup button does work but without any feedback given
https://www.reddit.com/r/toolbox/comments/fm1hvj/release_v530_coaching_chicken/fl1rkml/?context=3

Geo1088:
This should theoretically fix it, works on Chrome but my FF is being a bit wonky and I can't tell if it fixes or not there. Hopefully fixes #269.


creesch:
Tested it and it works, did add a little bit of extra feedback while we are at it.

Issue #95, opened 2019-06-13
mobile/small window tweaks


creesch:
Toolbox works on mobile Firefox and I often check some things on my phone. This made me aware of some things we probably can fox fairly easily without having to refactor much.

- popups in new modmail fall outside the window. This can be fixed by having css for a screenwidth under a certain amount of pixels that makes the popups act more like overlays by setting the left and right values to zero.
- When opening the new modmail user sidebar it doesn't have toolbox buttons. 
- popups can not be dragged as our dragging plugin doesn't work with touch.
- a full modbar becomes distorted.

From a quick test it seems that new modmail starts morphing to a mobile oriented layout from 768 pixels and lower, so that is what I'll target at as well I guess.

creesch:
Quick and dirty fixes for popups on smaller screens 

```css
@media screen and (max-width: 768px) {
  .mod-toolbox-rd .tb-popup {
    left: 0 !important;
    right: 0 !important;
    top: 0 !important;
    position: fixed !important;
  }

  .mod-toolbox-rd .tb-popup .tb-popup-content {
    min-width: 0 !important;
    max-height: calc(100vh - 105px) !important;
  }
  .mod-toolbox-rd #tb-toolbarshortcuts {
    display: none !important
  }
  .mod-toolbox-rd .utagger-type-list td {
    display: inline;
    height: 15px;
    line-height: 30px;
  }  
}
```

creesch:
Today I was reminded about this and did some updated work on the css

```css
@media screen and (max-width: 768px) {
    .mod-toolbox-rd.mod-toolbox-rd .tb-window-wrapper.tb-window-wrapper {
        top: 0;
        bottom: 0 !important;
        left: 0;
        right: 0;
        margin-bottom: 0;
        postion: fixed !important;
    }
    
    .mod-toolbox-rd .tb-profile-sidebar {
        display: none;
    }

    .mod-toolbox-rd .tb-profile-overlay .tb-window-tabs-wrapper {
        width: 100%;
        max-height: calc(100vh - 105px) !important;
        overflow: auto;
    }

    .mod-toolbox-rd .tb-page-overlay {
        padding: 0 !important;

    }
    .mod-toolbox-rd .tb-popup {
        left: 0 !important;
        right: 0 !important;
        top: 0 !important;
        position: fixed !important;
    }

    .mod-toolbox-rd .tb-popup .tb-popup-content {
        min-width: 0 !important;
[...]

PR #272, opened 2020-03-21
redirect from mobile to desktop

creesch:
There are a few things we can do to make toolbox function a bit better on mobile android. I am planning to tackle these in this PR. 

- [x] redirect mobile to desktop. It used to be that on mobile reddit you only had to set this once and it would remember, but this has been broken for a while. This fixes it from a toolbox setting by attempting to set the cookie for longer and redirecting automatically. 
- [ ] A variet of css fixes I have mention in #95

### DIFF
--- a/extension/data/background/handlers/mobilecookie.js
+++ b/extension/data/background/handlers/mobilecookie.js
@@ -1,0 +1,12 @@
+'use strict';
+
+messageHandlers.set('tb-mobile-cookie', (request, sender) => {
+    browser.cookies.set({
+        url: sender.tab.url,
+        path: '/',
+        domain: '.reddit.com',
+        name: 'mweb-no-redirect',
+        value: '1',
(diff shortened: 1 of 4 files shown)


Geo1088:
Would introduce a fair bit of new complexity, but would it be worth it to hide mobile-specific settings for desktop users?

creesch:
I don't think so, that's why I made it a general setting anyway so it is out of the way mostly.

creesch:
Considering that currently with fenix it is impossible to install any extension other than the ones mozilla whitelisted I am not going to continue work on it. 

There is kiwi browser which is chrome based and does support extensions but I haven't tested toolbox on there. There is also rumours that edge and vivaldi are looking into extension support on mobile (possibly based on the recently open sourced code of kiwi browser). I'll wait for developments there before I continue work here.

PR #282, opened 2020-03-25
Make ban and mute reason links clickable

creesch:
Pretty much as the title says. Accounts for both ban and mute reasons (thanks to @Blank-Cheque reminding me of mute reasons) and also does relative urls.


Geo1088:
Anything blobking merge?

PR #284, opened 2020-03-25
Fix {body} resolving to "> " for removal reasons

### LINKED ISSUE #283: Add a token for the removed selftext or body in mod macros
Please add a token for the removed selftext (submissions) or body (comments) in mod macros. There are some circumstances in which a post (generally a submission) is removed and we want to put the entire text that has been removed into the removal message (as a courtesy for anyone currently reading the post).

`{body}` would be consistent with the placeholder that AutoModerator uses for this.

Blank-Cheque:
This was due to TBCore.getApiThingInfo not accounting for the fact that the content of a submission vs a comment is conveyed with different JSON keys. 

closes #283

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -1068,7 +1068,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                     const data = response.data;
 
                     let user = data.children[0].data.author;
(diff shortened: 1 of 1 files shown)


creesch:
Looks good to me at first glance

PR #288, opened 2020-03-28
feat: Macro per type

### LINKED ISSUE #37: Add option to make macros post, comment or modmail specific.
Probably just add checkbox for each instance where if none are checked it will assume it needs to show on all. 
Can probably be added without a schema change if we add it like that.

SpyTec:
Closes #37

Adds context options for creating and editing macros. By default it will all be false for both adding and editing.
image

Existing macros which have not been edited to have these new fields will show on everything. 
image

### DIFF
--- a/extension/data/modules/macros.js
+++ b/extension/data/modules/macros.js
@@ -137,7 +137,7 @@ function modmacros () {
 
                     // if we don't have a config, get it.  If it fails, return.
                     getConfig(info.subreddit, (success, config) => {
-                    // if we're a mod, add macros to top level reply button.
+                        // if we're a mod, add macros to top level reply button.
                         if (success && config.length > 0) {
                             const $tbUsertextButtons = $thing.find('.usertext-buttons .tb-usertext-buttons'),
                                   macroButtonHtml = `<select class="tb-macro-select tb-action-button" data-subreddit="${info.subreddit}"><option value=${MACROS}>macros</option></select>`;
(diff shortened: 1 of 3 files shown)


creesch:
Don't see any obvious issues with this. I'll do some actual testing later today and then merge.

Geo1088:
Conflict caused by merging #297, resolving...

Geo1088:
Hmm, actually, I take back my approval. I'm having trouble getting a macro with "Modmail" context to show up in the dropdown in modmail threads.

extension/data/modules/macros.js
```diff
@@ -163,7 +178,7 @@ function modmacros () {
                     const macroButtonHtml = `<select class="tb-macro-select tb-action-button" data-subreddit="${info.subreddit}"><option value=${MACROS}>macros</option></select>`;
                     $body.find('.ThreadViewerReplyForm__replyOptions').after(`<div class="tb-usertext-buttons tb-macro-newmm">${macroButtonHtml}</div>`);
 
-                    populateSelect('.tb-macro-select', info.subreddit, config);
+                    populateSelect('.tb-macro-select', info.subreddit, config, 'comment');
```
Geo1088, line 181: Unless I'm misunderstanding something, this line requires macros to have the "comment" type to work in new modmail, which seems wrong. Changing it to "modmail" fixes it in the case I was testing, at least. Would like @SpyTec to check my logic here though.

```suggestion
                    populateSelect('.tb-macro-select', info.subreddit, config, 'modmail');
```

SpyTec: Good catch, double checked and you're right

Issue #289, opened 2020-03-29
Go over caching one more time


creesch:
Because it seems like I introduced some oddities last time I reworked caching. 

- [This isn't actually consumed on the front](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/background/handlers/cache.js#L42-L51) so cache objects aren't cleared for existing pages. I am not sure if that is a good or a bad thing though, certainly something that needs to figure out. 
- Some cached data is set on init if the cache object is empty, like moddedsubs. However if that is the first tab to open (and thus init of the background page) [this function](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/background/handlers/cache.js#L20-L51) simply clears everything.

creesch:
A little update... seems like some things got conflated with each other... https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbcore.js#L1469

PR #290, opened 2020-03-29
Fix caching timeout mechanism

creesch:
Goes over the things I raised in #289 

Still needs some testing to see if I didn't break stuff but this does actually clear the cache locally as well when needed and makes sure already fetched subreddit data doesn't go to waste.

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -135,46 +135,8 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
 
         TBCore.modsSub = subreddit => TBCore.mySubs.includes(subreddit);
 
-        if (newModSubs && newModSubs.length > 0) {
-            TBCore.mySubs = [];
-            TBCore.mySubsData = [];
-            $(newModSubs).each(function () {
-                const sub = this.data.display_name.trim();
-                if (!TBCore.modsSub(sub)) {
-                    TBCore.mySubs.push(sub);
-                }
-
-                let isinthere = false;
-                $(TBCore.mySubsData).each(function () {
-                    if (this.subreddit === sub) {
-                        isinthere = true;
-                    }
-                });
-
-                if (!isinthere) {
-                    const subredditData = {
-                        subreddit: sub,
-                        subscribers: this.data.subscribers,
-                        over18: this.data.over18,
-                        created_utc: this.data.created_utc,
-                        subreddit_type: this.data.subreddit_type,
-                        submission_type: this.data.submission_type,
-                        is_enrolled_in_new_modmail: this.data.is_enrolled_in_new_modmail,
-                    };
-
-                    TBCore.mySubsData.push(subredditData);
-                }
-            });
-
-            TBCore.mySubs = TBHelpers.saneSort(TBCore.mySubs);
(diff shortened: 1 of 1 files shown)


creesch:
Okay, seems to break some buttons on initial load. Going to need to rework it a bit.

extension/data/tbcore.js
```diff
@@ -258,7 +215,54 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
         }
         </style>`);
 
+        TBCore.modsSub = subreddit => TBCore.mySubs.includes(subreddit);
```
Geo1088, line 218: Any reason all this was moved down from where it was before?

extension/data/tbcore.js
```diff
@@ -258,7 +215,54 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
...
+                if (!TBCore.modsSub(sub)) {
+                    TBCore.mySubs.push(sub);
+                }
+
+                let isinthere = false;
+                TBCore.mySubsData.forEach(tbCoreSubData => {
+                    if (tbCoreSubData.subreddit === sub) {
+                        isinthere = true;
+                    }
+                });
```
Geo1088, line 235: Can shorten this section with `.some`:

```js
TBCore.mySubsData.some(tbCoreSubData => tbCoreSubData.subreddit === sub);
```

extension/data/tbcore.js
```diff
@@ -1459,14 +1463,15 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
...
+        let firstCacheTimeout = true;
         // Listen to background page communication and act based on that.
         browser.runtime.onMessage.addListener(message => {
             switch (message.action) {
             case 'clearCache': {
                 TBCore.clearCache(true);
                 break;
             }
-            case 'tb-settings-update': {
+            case 'tb-cache-timeout': {
```
Geo1088, line 1474: Not sure off the top of my head, what's the result of changing this to listen for a different event?

creesch: The first event didn't exist... basically it wasn't working before.

extension/data/tbcore.js
creesch: I honestly don't remember, I had to move a bunch of stuff so it might have been the result of that. I also might have thought it fitted better at this place.

Geo1088: No real issue other than the big diff, should be fine.

extension/data/tbcore.js
Geo1088: Ah, lovely. Makes sense then.

extension/data/tbcore.js
creesch: I might do that, as you'll note though this is existing code ;)

Geo1088: Not quite, you did reformat it from `$().each()` already :P

creesch: Well yeah but... that was something that was supposed to be refactored already by someone else :P 

I am mostly being lazy here now :P

Geo1088, line 235:
```suggestion
                const isinthere = TBCore.mySubsData.some(tbCoreSubData => tbCoreSubData.subreddit === sub);
```

PR #291, opened 2020-03-29
feat(RR): Add auto-archive send as modmail

### LINKED ISSUE #90: Feature request: add a setting to automatically archive comment removal modmail messages
This was requested in /r/toolbox [here](https://www.reddit.com/r/toolbox/comments/5ecqfk/feature_request_the_removal_reasons_prompt_should/) and [here](https://www.reddit.com/r/toolbox/comments/5t2t8q/feature_request_automatically_archive_modmail/).

The "Send removal reasons as subreddit" setting currently warns "Note: these will appear in modmail and potentially clutter it up.", which feature would help avoid.

A couple questions:
  - Is it possible to enforce this setting (and the send as sub setting) for all mods in a subreddit?
  - If a subreddit is still using the old mod mail, is it better to have this setting be a noop or to hit the /archive endpoint anyway? The endpoint returns 404 if the conversationId is invalid.
  - Does toolbox currently contain any util for determining if a subreddit is on the new mod mail? (Aside from checking for the `mod.reddit.com` hostname, which only works when the user is checking modmail)

I'm happy to open a PR for this feature.

### LINKED ISSUE #117: Ability to archive modmail send through toolbox
For use in modbutton and for use in removal reasons.

Needs to make use of the new modmail api, which also means we need to check the subreddit data for `is_enrolled_in_new_modmail` first. Probably before we offer the option to archive in the first place.

### LINKED ISSUE #119: Perform schema version checks on sub config
For some reason we don't right now. That's not great.

SpyTec:
Closes #117, Closes #90

This introduces auto-archive option in removal reasons when sending as PM through new modmail. Default setting was also added to subreddit settings but not user settings.

Not been able to test it fully due to not having any test accounts at hand

image

### DIFF
--- a/extension/data/modules/config.js
+++ b/extension/data/modules/config.js
@@ -131,7 +131,7 @@ function tbconfig () {
                                     </li>
                                     <li>
                                         <input class="reason-auto-archive" type="checkbox" id="type-auto-archive" ${configData.removalReasons.autoArchive ? 'checked' : ''}/>
-                                        <label for="type-auto-archive">Auto-archive sent modmail pm <b>Note</b>: Only works on new modmail.</label>
+                                        <label for="type-auto-archive">Auto-archive sent modmail pm <b>Note:</b> Only works on new modmail.</label>
                                     </li>
                                 </ul>
                             </li>
--- a/extension/data/modules/config.js
+++ b/extension/data/modules/config.js
@@ -129,6 +129,10 @@ function tbconfig () {
                                         <input class="reason-as-sub" type="checkbox" id="type-as-sub" ${configData.removalReasons.typeAsSub ? 'checked' : ''}/>
                                         <label for="type-as-sub">Send pm via modmail as subreddit <b>Note:</b> This will clutter up modmail.</label>
                                     </li>
+                                    <li>
+                                        <input class="reason-auto-archive" type="checkbox" id="type-auto-archive" ${configData.removalReasons.autoArchive ? 'checked' : ''}/>
+                                        <label for="type-auto-archive">Auto-archive sent modmail pm.</label>
+                                    </li>
                                 </ul>
                             </li>
                             <li>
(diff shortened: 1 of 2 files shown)


extension/data/modules/config.js
```diff
@@ -129,6 +129,10 @@ function tbconfig () {
                                         <input class="reason-as-sub" type="checkbox" id="type-as-sub" ${configData.removalReasons.typeAsSub ? 'checked' : ''}/>
                                         <label for="type-as-sub">Send pm via modmail as subreddit <b>Note:</b> This will clutter up modmail.</label>
                                     </li>
+                                    <li>
+                                        <input class="reason-auto-archive" type="checkbox" id="type-auto-archive" ${configData.removalReasons.autoArchive ? 'checked' : ''}/>
+                                        <label for="type-auto-archive">Auto-archive sent modmail pm.</label>
```
creesch: Possibly add "**Note:** Only works on new modmail"

extension/data/modules/removalreasons.js
```diff
@@ -807,9 +815,10 @@ function removalreasons () {
                 if (typeof logLink !== 'undefined') {
                     reason = reason.replace('{loglink}', logLink);
                 }
-
-                const notifyByPM = notifyBy === 'pm' || notifyBy === 'both',
-                      notifyByReply = notifyBy === 'reply' || notifyBy === 'both';
+                const subredditData = TBCore.mySubsData.find(s => s.subreddit === data.subreddit),
```
creesch: It's always possible this has expired and thus empty so you want to wrap it in a `TBCore.getModSubs()` callback

rahulthewall:
@creesch Do you have any idea when this feature would be available as a release?

creesch:
We don't have a set release date yet, as it stands we still have some testing to do and due to circumstances development is a bit slower as we'd like.

rahulthewall:
Thanks for the information @creesch, and for the hard work that you've put in this tool. 

This feature would solve a lot of issues for us, and I look forward to using it when you finally release it.

PR #299, opened 2020-04-19
NewModMailPro: Feature to check for new activity before submitting a reply

molenzwiebel:
This PR adds a new feature to NewModMailPro that polls the server for new activity on a modmail thread before submitting a reply. This should hopefully lead to less issues where two moderators respond to the same thread at the same time.

I've tested the feature in numerous settings and it should be stable. It is compatible with modmail macros (it will check for new activity when posting a macro) and the reply type check (it will check for reply type before polling server). I had to slightly change the modmail preview since it previously removed the preview on button press, instead of on form submission.

Some things that should probably be reviewed:
- Should the feature be enabled by default?
- Should the modal be positioned at the submit button or at the center-top of the page?

Documentation and ESLint should check out.

### DIFF
--- a/extension/data/background/handlers/webrequest.js
+++ b/extension/data/background/handlers/webrequest.js
@@ -47,7 +47,10 @@ async function getOAuthTokens (tries = 1) {
     // be able to read the new cookie and all is well, without having to deal with
     // the Reddit OAuth flow ourselves.
     if (tries < 3) {
-        await makeRequest('https://mod.reddit.com/mail/all');
+        await makeRequest({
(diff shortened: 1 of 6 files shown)


OldGoldie:
Do you think this could also be an option to check for an exist removal reason?

May be a bit hard because not all subs would sticky the removal reason etc. so there would have to be some logic to determine if there is one. Just throwing the idea.

creesch:
That would fall well outside this specific PR. Feel free to make an issue for it though.

Issue #261, opened 2020-03-14
modmail preview: fix relative urls


creesch:
When typing /r/sub or /u/user clicking on them doesn't work as we are in the `mod` subdomain.

Geo1088:
This will be a bit of a pain to work on because Snuownd doesn't have any flag or option for outputting fully-qualified URLs for these. Our options are either to fork Snuownd and add that functionality ourselves, or write some jank regex substitution to add `https://www.reddit.com` to the beginning of all generated `href`s that start with `/`.

Geo1088:
We could probably fix this for 5.3.1 if we just do it the jank way, `.replace('href="/r/', 'href="https://www.reddit.com/r/').replace('href="/u/', 'href="https://www.reddit.com/user/')`

Thoughts?

SpyTec:
Not really a jank way, but yeah. Simple regex replace would suffice

creesch:
afaik you can't type relative urls from modmail normally so we can just replace ` href="/`, right?

creesch:
though people can manually I suppose...

Geo1088:
Actually yeah, that does make sense, just `.replace(/href="/g, 'href="https://www.reddit.com/')` should do the trick. Relative URLs I don't think we need to worry about.

SpyTec:
Right it's relative URLs.. Reddit makes these links full-links after submission

![image](https://user-images.githubusercontent.com/6598829/77259080-4e635300-6c7f-11ea-9969-8a6fc998261e.png)

Issue #301, opened 2020-04-20
Modmail preview displays < > as &lt; &gt;


Geo1088:
likely related #280

creesch:
Oh that might be the result of purify and the markdown render thing not playing nice...

PR #302, opened 2020-04-20
Modmail preview fixes

### LINKED ISSUE #280: Some HTML tags aren't displayed in Modmail Preview
As the title says, some HTML tags (and things starting with `<` in general) aren't displayed in the Modmail Preview.

screenshot

creesch:
So far fixes Fix #280 and #301 (dunno why I thought it was 330 before..)

Slightly less secure I suppose as we render first then purify it but the other way around makes things go wonky as seen in the two linked issues.


creesch:
Also fixes #261 now.

Issue #118, opened 2019-07-09
Replace RR HTML parsing with replaced tokens


Geo1088:
Instead of using stuff like `<textarea>`, `<select>`, `<input>` inside removal reason bodies to get custom text input, we should have tokens that are replaced with the appropriate elements without having to parse HTML. For example:

- `<input type="text">` ~> `{textfield}`
- `<textarea>` ~> `{textarea}`
- `<select>` ~> `{option|other option|etc}`

things we'll have to do

- [ ] new schema version, remove html parsing
- [ ] write conversion from old to new (make sense of as much old html as we can)
- [ ] implement syntax for the multiple-choice one - will probably require regex parsing, and should also support escaping the inner pipes and right-brace if possible so those can be included literally in options

SpyTec:
I would suggest to use `|` as delimiter as it's seldom used in comments/posts.

And for templating, I'd suggest to mix two chars for start and finish, or use them twice. 
```
{$ thing here $}
{% thing here %}
{{ thing here }} <-- I prefer this one
```

This is so you can use the characters in the comments without running into issues so easily. And we can make sure people can fix it if they want by escaping

```
{{select
|I am an option
|I'm \{\{ enclosed }}
}}
```

If we ever introduce parameters for types, we can do something like 
```
{{switch;name:fancyname|Switch label here}}
{{select;typeahead;hidden-by:fancyname|foo|bar|baz}}
```

creesch:
Just going over issues at the moment and wanted to note that: 

> I would suggest to use | as delimiter as it's seldom used in comments/posts.

Probably isn't true as I have seen it used in footers a bunch to put in a bunch of links to stuff like rules, wikis, etc. 

Not sure what the best approach here is except possibly building in escape functionality for whatever delimiter character we decide on.

SpyTec:
So, instead of doing tokens. Another option would be to keep HTML parsing but utilize web-components (if that's possible).

* `<input type="text">` ~> `<tb-text>`
* `<textarea>` ~> `<tb-text multiline>`
* `<select>` ~> `<tb-select><tb-option option></tb-option></tb-select>`

This means we can 
* filter out everything that isn't our components,
* associate toolbox info for each component, and
* still parse HTML to get attributes in an easy way

PR #305, opened 2020-04-27
Support placeholder, label, value attributes in removal reasons

molenzwiebel:
As per [this](https://www.reddit.com/r/toolbox/comments/g8zjfx/feature_request_support_for_label_attribute_for/) feature request on the sub, this PR adds support for the `placeholder`, `label` and `value` attributes in removal reasons.

The linked post explains the reasoning behind allowing the attributes. Internally, the default removal reason already had a placeholder but it was never parsed because it was missing from the whitelist.

I've confirmed that the `label`/`value` attribute on `<option>` tags works as expected. I've also confirmed that an empty textarea with a placeholder will not take that placeholder as value or other weirdness.

### DIFF
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -394,7 +394,7 @@ function removalreasons () {
 
                     // Set up markdown renderer
                     SnuOwnd.DEFAULT_HTML_ELEMENT_WHITELIST.push('select', 'option', 'textarea', 'input');
-                    SnuOwnd.DEFAULT_HTML_ATTR_WHITELIST.push('id');
+                    SnuOwnd.DEFAULT_HTML_ATTR_WHITELIST.push('id', 'placeholder', 'label', 'value');
                     const parser = SnuOwnd.getParser(SnuOwnd.getRedditRenderer(SnuOwnd.DEFAULT_BODY_FLAGS | SnuOwnd.HTML_ALLOW_ELEMENT_WHITELIST));
 
                     // Render header and footer
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -685,7 +685,9 @@ function removalreasons () {
                 // Get input from HTML-formatted reason
                 const htmlReason = $this.find('.reason-content');
                 htmlReason.find('select, input, textarea').each(function () {
-                    customInput.push(this.value);
+                    // Value, if it is not empty. If it is, placeholder.
+                    // If no placeholder, empty string.
+                    customInput.push(this.value || this.placeholder || '');
                 });
 
                 // Get flair data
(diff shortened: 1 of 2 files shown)


creesch:
Oh boy, we have been holding back on removal reason html parsing things due to #118 probably okay here I suppose but good to know for the future. 

@Geo1088 what do you think?

Geo1088:
Was mostly holding back from implementing the placeholder thing just because I didn't want to put in the effort before tackling #118. If somebody else has already written it in though, I don't see an issue with including it now, especially since we're clearly in no rush to take care of #118.

Also related: https://www.reddit.com/r/toolbox/comments/er30rx/issues_with_placeholder_text_in_multiline_inputs/

Geo1088:
Is this PR tested? I don't see anything here that would make `placeholder` work as expected, since the value of a blank `<textarea placeholder="...">` is still an empty string IIRC.

molenzwiebel:
As per my initial message, 

> I've also confirmed that an empty textarea with a placeholder will not take that placeholder as value

If you want to be consistent with the entirety of the map, then placeholders should act as hints for the user and not as default values. With that in mind, treating empty textareas as empty even if they have a placeholder attribute seems to be the right call.

Geo1088:
Must've missed that, my bad. The feature request I linked asks for `placeholder` to be used as a default value used if no text is typed, which seems like a reasonable request to me. I'm not sure what good whitelisting it does otherwise - it'd be a hint for what to type there, but isn't that normally evident by the rest of the message?

I know this doesn't really fit the semantic purpose of `placeholder`, but I can't think of a better way to handle it here.

molenzwiebel:
It'd be fairly easy to add, but I'm worried it may cause confusion if gray text people are trained to see as suggestion/placeholder suddenly gets posted as part of the message. 

One idea could be to use css to make the placeholder have the same color as normal text. That way we don't need to manually implement placeholder logic and at the same time should cause less confusion for users.

Should only be a couple lines of css/code to get that working.

Geo1088:
Yeah that seems reasonable to me. This should do it as far as CSS goes, ensuring the color is consistent with RES nightmode and working with `<textarea>` as well as `<input type="text">`:

```css
.mod-toolbox-rd .reason-popup ::placeholder {
  color: inherit;
}
```

creesch:
A bit more complex but would also allow people to edit the placeholder is to simply copy the placeholder to value. That way we don't have to rely on CSS (and possibly subreddit CSS fucking with it accidentially).

molenzwiebel:
The problem with copying the placeholder to the value is that it becomes ambiguous when it should be reset to the placeholder. If we reset it when the field is empty, it'll interfere with people that want a different message entirely. If we reset it on page load, it'll interfere with the persisting of the contents, etc, etc.

I've committed the placeholder hack for now. Looks like this:

image

PR #317, opened 2020-05-21
Add usernotes manager icon to "moderated subreddits" popup

### LINKED ISSUE #316: Add icon for user notes to "moderated subreddits"
It's a little confusing that the only place to access the full user notes page is under the other :arrow_forward: menu which I normally consider to be the "user profile menu". It seems like it should just be next to all of the other subreddit-specific stuff in the "moderated subreddits" menu.

Thanks.

Geo1088:
Fixes #316.

### DIFF
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -302,7 +302,7 @@ function modbar () {
                     const $this = $(this);
                     if (!$this.hasClass('tb-mysubs-activated')) {
                         $this.addClass('tb-mysubs-activated');
-                        const $popup = TB.ui.popup({
+                        TB.ui.popup({
                             title: 'Subreddits you moderate',
                             tabs: [
                                 {
@@ -321,7 +321,6 @@ function modbar () {
                         });
                         // Focus the filter bar for convenience
                         $('#tb-livefilter-input').focus();
-                        // Add event listener to open usernotes manager
                     } else {
                         $this.removeClass('tb-mysubs-activated');
                         $('.subreddits-you-mod-popup').remove();
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -264,6 +264,7 @@ function modbar () {
             let subList = '',
                 livefilterCount;
             const configEnabled = TB.storage.getSetting('TBConfig', 'enabled', false),
+                  usernotesEnabled = TB.storage.getSetting('UserNotes', 'enabled', false),
                   subredditColorSalt = self.setting('subredditColorSalt');
             TBCore.getModSubs(() => {
                 self.log('got mod subs');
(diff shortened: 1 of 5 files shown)


extension/data/modules/modbar.js
```diff
@@ -264,26 +264,28 @@ function modbar () {
...
-        <a title="/r/${this.subreddit} traffic stats" target="_blank" href="${TBCore.link(`/r/${this.subreddit}/about/traffic`)}" data-type="traffic" data-subreddit="${this.subreddit}" class="tb-icons">${TBui.icons.subTraffic}</a>
-        ${configEnabled ? `<a title="/r/${this.subreddit} config" href="javascript:;" class="tb-config-link tb-icons" data-subreddit="${this.subreddit}">${TBui.icons.tbSubConfig}</a>` : ''}
-    </td>
-</tr>
-`;
+                    subList += `
+                        <tr style="border-left: solid 3px ${subColor} !important;" data-subreddit="${this.subreddit}">
+                            <td class="tb-my-subreddits-name"><a href="${TBCore.link(`/r/${this.subreddit}`)}" target="_blank">/r/${this.subreddit}</a></td>
+                            <td class="tb-my-subreddits-subreddit">
+                                <a title="/r/${this.subreddit} modmail!" target="_blank" href="${TBCore.link(`/r/${this.subreddit}/message/moderator`)}" data-type="modmail" data-subreddit="${this.subreddit}" class="tb-icons">${TBui.icons.oldModmail}</a>
```
SpyTec, line 279: Modmail is the only one that has exclamation point in the title attribute. Not related to PR though

extension/data/modules/modbar.js
```diff
@@ -264,26 +264,28 @@ function modbar () {
...
+                                <a title="/r/${this.subreddit} unmoderated" target="_blank" href="${TBCore.link(`/r/${this.subreddit}/about/unmoderated`)}" data-type="unmoderated" data-subreddit="${this.subreddit}" class="tb-icons">${TBui.icons.unmoderated}</a>
+                                <a title="/r/${this.subreddit} moderation log" target="_blank" href="${TBCore.link(`/r/${this.subreddit}/about/log`)}" data-type="modlog" data-subreddit="${this.subreddit}" class="tb-icons">${TBui.icons.modlog}</a>
+                                <a title="/r/${this.subreddit} traffic stats" target="_blank" href="${TBCore.link(`/r/${this.subreddit}/about/traffic`)}" data-type="traffic" data-subreddit="${this.subreddit}" class="tb-icons">${TBui.icons.subTraffic}</a>
+                                ${usernotesEnabled ? `<a title="/r/${this.subreddit} usernotes" href="javascript:;" class="tb-un-config-link tb-icons" data-subreddit="${this.subreddit}">${TBui.icons.usernote}</a>` : ''}
```
SpyTec, line 284: What's with using `tb-un-config-link` class?

extension/data/modules/modbar.js
```diff
@@ -319,6 +321,7 @@ function modbar () {
                         });
                         // Focus the filter bar for convenience
                         $('#tb-livefilter-input').focus();
+                        // Add event listener to open usernotes manager
```
SpyTec, line 324: Where is the event listener?

extension/data/modules/modbar.js
Geo1088: The usernotes manager doesn't need its own listener because it uses the listener already registered by the usernotes module. This class is what the usernotes module listens for clicks on to open the manager.

extension/data/modules/modbar.js
Geo1088: Whoops, accidentally left that from local copy when I thought I'd need to add another listener to the code here. Will remove.

PR #318, opened 2020-05-21
Make removal reasons work when removing comments through context popup

### LINKED ISSUE #313: Removal reasons don't get triggered when removing from within context popup
Likely need to change the RR trigger selector to also match the synthetic remove buttons we generate for the synthetic listing displays.

Geo1088:
Fixes #313.


creesch:
Looks good to me, as spytec also signed of I am merging this one.

PR #327, opened 2020-06-08
Refactoring and unify window/popup code

### LINKED ISSUE #240: Unify markup/classes for dialogs/windows/notifications
As it is now, all three of these are basically the same thing visually with slight differences:

- Windows are popups with an overlay background and a larger header/tab size
- Notifications are popups with `position: static`, no draggability, and no footer

By combining these to use the same base class with modifying classes for their differences, we can clean up a lot of the existing CSS. They already share [a lot of the same code](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/styles/tbui.css#L1-L94).

I propose we use `tb-window` as the base class for these, `tb-window-free` as a modifier for free (draggable) popups, and `tb-window-large` as a modifier for windows with larger headers, tabs, and footers. This would mean:

- `TBui.popup` would return a `.tb-window.tb-window-free` for a small, draggable window, 
- `TBui.overlay` would use a `.tb-window.tb-window-large` for a large, non-draggable window, and
- `TBui.showNotification` would use a `.tb-window` with no modifier classes for a small, static, non-draggable window.

Classes for sub-elements of these structures (`tb-popup-content`, etc.) would be merged into single `tb-window-*` classes, and any styles that need to apply to only a single class of window can be refactored around that.

~~Additionally, currently there are more or less two types of draggable window: absolutely positioned (move with page scroll, e.g. usernotes) and fixed positioned (don't move on scroll, e.g. personal notes). We should eliminate one or the other for a more consistent user experience, and I'm leaning towards keeping everything fixed-position to avoid issues with users losing windows by scrolling around.~~ Deemed not worth it for the small amount of complexity this adds, and is mostly a matter of preference.

Also, tab display for popups and overlays is currently different, with the latter defaulting to displaying tabs as a vertical list on the left of the window. To implement this proposal, the default tab arrangement for overlays will have to be changed to horizontal to match popups, and an additional `tb-window-vertical-tabs` helper class can be added to deal with areas where vertical tabs aready exist. This will mean that popups will have access to vertical tabs natively, which is cool because it means we'll be able to use those for stuff like personal notes in the future.

Geo1088:
Fixes #240.

Honestly don't remember what's in this branch right now but I know it breaks things. Will come back to this.

### DIFF
--- a/extension/data/modules/comment.js
+++ b/extension/data/modules/comment.js
@@ -525,10 +525,6 @@ function comments () {
                     TBui.tbRedditEvent($comments);
                     $('time.timeago').timeago();
                     $comments.find(`.tb-thing[data-comment-id="${commentID}"] > .tb-comment-entry`).css('background-color', '#fff8d5');
-                    // Close the popup
-                    $contextPopup.on('click', '.close', () => {
-                        $contextPopup.remove();
-                    });
                 });
             });
         }
--- a/extension/data/modules/devtools.js
+++ b/extension/data/modules/devtools.js
@@ -67,11 +67,6 @@ function devtools () {
                         top: event.pageY - 10,
                         display: 'block',
                     });
-
-                // Close the popup
-                $apiPopup.on('click', '.close', () => {
-                    $apiPopup.remove();
-                });
             });
         }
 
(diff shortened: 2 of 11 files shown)


Geo1088:
been too long since I did this and I committed in an incomplete broken state so not worth picking this back up, will make a new branch off current master and restart this

PR #335, opened 2020-07-02
recognize toolbox-generated comments as comments for RRs

### LINKED ISSUE #334: RR triggered when removing comment in context-popup with comment RR disabled for both toolbox and subreddit.
The same comment didn't trigger RR when removed in the regular context.

Geo1088:
Fixes #334


creesch:
I think we might need some extra code to also make this work on redesign where the popup can also show up.

Geo1088:
ugh right because this is in an `if (isOldReddit)`, how did I miss that... Gimme a bit

Geo1088:
*20 days later*

Hacked on another condition to handle our markup the same way as old Reddit even if we're on new. A better way to handle this for the future would be to emit faked jsAPI events for markup we generate, but that's gonna be big and out of scope for this bugfix.

PR #336, opened 2020-07-02
Button to display message source in the New Modmail

### LINKED ISSUE #323: Add "source" button to new modmail
Please add a `source` button to each modmail to view the Markdown source of the message.

Thanks!

P.S. I realize the `source` button that we all know and love is a RES feature, but [it doesn't look like RES is going to be supporting this feature on the redesign in general any time soon](https://github.com/honestbleeps/Reddit-Enhancement-Suite/issues/4951), let alone new modmail. And while I can almost entirely avoid the redesign as a moderator, I cannot avoid new modmail.

Xeoth:
If I did everything right, this fixes #323

### DIFF
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -406,8 +406,6 @@ function newmodmailpro () {
 
                 const $target = $(event.target);
 
-                // console.log();
-
                 $target.closest('.Thread__message').append('<button class="tb-source-button tb-action-button">Toggle Source</button>');
 
                 // Fetch and store the conversation info in cache
@@ -415,6 +413,8 @@ function newmodmailpro () {
                 TBStorage.setCache('NewModmailPro', 'current-conversation', await TBApi.apiOauthGET(`/api/mod/conversations/${currentID}`).then(r => r.json()));
 
                 $body.on('click', '.tb-source-button', async e => {
+                    console.log(e);
+                    // console.log($(e.target))
                     const $currentSourceBtn = $(e.currentTarget.parentElement);
 
                     // Getting the ID of the message on which the button was clicked
@@ -422,16 +422,10 @@ function newmodmailpro () {
                     const $currentSourceField = $(`#tb-source-${activeMessageID}`);
 
                     // Toggling the source
-                    if ($currentSourceField && $currentSourceField.is(':visible')) {
-                        // If the source is visible, hide it
+                    if ($currentSourceField.length) {
+                        // If the source field exists, toggle it
 
-                        $currentSourceField.hide();
-                        return;
-                    } else if ($currentSourceField && $currentSourceField.is(':hidden')) {
-                        // If the source field exists (has already been requested), but has been hidden, just show it
-
-                        $currentSourceField.show();
-                        return;
+                        $currentSourceField.toggle();
                     } else {
                         // If the source field is not present (has not been requested yet), request it and create
                         // a div+textarea with the source.
@@ -440,11 +434,11 @@ function newmodmailpro () {
                         const conversationInfo = await TBStorage.getCache('NewModmailPro', 'current-conversation');
                         const messageSource = conversationInfo.messages[activeMessageID].bodyMarkdown;
 
-                        $currentSourceBtn.closest('div.Thread__message').append(`
-                            <div class="tb-source-field" id="tb-source-${activeMessageID}">
-                                <textarea readonly>${messageSource}</textarea>
-                            </div>
-                        `);
+                        $currentSourceBtn.closest('.Thread__message').append(`
+                        <div class="tb-source-field" id="tb-source-${activeMessageID}">
+                            <textarea readonly></textarea>
+                        </div>`);
+                        $(`#tb-source-${activeMessageID} textarea`).text(messageSource);
                     }
                 });
             });
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -412,9 +412,7 @@ function newmodmailpro () {
                 const currentID = event.detail.data.post.id;
                 TBStorage.setCache('NewModmailPro', 'current-conversation', await TBApi.apiOauthGET(`/api/mod/conversations/${currentID}`).then(r => r.json()));
 
-                $body.on('click', '.tb-source-button', async e => {
-                    console.log(e);
-                    // console.log($(e.target))
+                $('.tb-source-button').unbind().click(async e => {
                     const $currentSourceBtn = $(e.currentTarget.parentElement);
 
                     // Getting the ID of the message on which the button was clicked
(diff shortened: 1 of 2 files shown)


extension/data/modules/newmodmailpro.js
```diff
@@ -390,6 +390,40 @@ function newmodmailpro () {
...
+
+        // 'Source' button to display message source
+        window.addEventListener('TBNewPage', async event => {
+            if (event.detail.pageType === 'modmailConversation') {
+                // Add the button to the buttons over the message
+                $('.Thread__message').append('<button class="tb-source-button tb-action-button">Show Source</button>');
+
+                // Fetch and store the conversation info in cache
+                const currentID = event.detail.pageDetails.conversationID;
+                TBStorage.setCache('NewModmailPro', 'current-conversation', await TBApi.apiOauthGET(`/api/mod/conversations/${currentID}`).then(r => r.json()));
```
creesch, line 402: I am not sure we need to do the request for every modmail loaded. Possibly just do the request the first time source is clicked

extension/data/modules/newmodmailpro.js
```diff
@@ -390,6 +390,40 @@ function newmodmailpro () {
                 }
             });
         }
+
+        // 'Source' button to display message source
+        window.addEventListener('TBNewPage', async event => {
+            if (event.detail.pageType === 'modmailConversation') {
+                // Add the button to the buttons over the message
+                $('.Thread__message').append('<button class="tb-source-button tb-action-button">Show Source</button>');
```
creesch, line 398: You also have conversations with collapsed messages for which you need to account. We already do this elsewhere in toolbox, I'll look it up in a bit.

Xeoth:
Thanks for your comments, Creesch. I wouldn't have noticed those. I fixed what needed fixing and added a toggling feature. It should all be good now.

Please review and leave a comment if you find something that still needs a fix.

extension/data/modules/newmodmailpro.js
```diff
@@ -390,6 +397,52 @@ function newmodmailpro () {
...
+
+                const $target = $(event.target);
+
+                $target.closest('.Thread__message').append('<button class="tb-source-button tb-action-button">Toggle Source</button>');
+
+                // Fetch and store the conversation info in cache
+                const currentID = event.detail.data.post.id;
+                TBStorage.setCache('NewModmailPro', 'current-conversation', await TBApi.apiOauthGET(`/api/mod/conversations/${currentID}`).then(r => r.json()));
+
+                $('.tb-source-button').unbind().click(async e => {
```
creesch, line 415: `unbind` is actually deprecated. You probably found it googling as the listener is triggered multiple times. A different solution would be to use `e.stopImmediatePropagation()`

extension/data/modules/newmodmailpro.js
creesch: You haven't actually resolved this, for every message in a conversation you are still doing a request. 

My suggestion would be to set up your cache something along these lines: 

1. Keep a local boolean named `conversationCached` and set it to `false` by default.
2. Attach the buttons to the replies.
3. When a source button is clicked check the boolean, if true use cache for the source if false fetch the conversation and set the cache. 
4. Clear the boolean based on a `TBNewPage`  event.

extension/data/modules/newmodmailpro.js
```diff
@@ -390,6 +397,52 @@ function newmodmailpro () {
...
+                    const $currentSourceBtn = $(e.currentTarget.parentElement);
+
+                    // Getting the ID of the message on which the button was clicked
+                    const [activeMessageID] = $currentSourceBtn.find('.Message__date')[0].pathname.split('/').slice(-1);
```
creesch, line 419: You already have the id from the jsAPI event, so this isn't really needed.

extension/data/modules/newmodmailpro.js
```diff
@@ -390,6 +397,52 @@ function newmodmailpro () {
...
+
+        if (sourceButton) {
+            TB.listener.on('author', async event => {
+                if (event.detail.type !== 'TBmodmailCommentAuthor') {
+                    return;
+                }
+
+                const $target = $(event.target);
+
+                $target.closest('.Thread__message').append('<button class="tb-source-button tb-action-button">Toggle Source</button>');
```
creesch, line 409: I really don't like the placement of this button, my suggestion would be to add it to the message header and change the class to `tb-general-button`. 

Like so 

image

extension/data/styles/newmodmailpro.css
```diff
@@ -36,6 +36,22 @@
     margin-bottom: 15px;
 }
 
+/* source textarea
+*/
+
+.tb-source-button {
```
creesch, line 42: Following my other comment this margin needs to change so it fits in the header.

The quote and Report button both have a left margin of 16px 

```css
body.mod-toolbox-rd .tb-source-button {
    margin-left: 16px;
}
```

creesch (changes requested):
So uh yeah I ended up testing my suggested changes to make sure that what I was saying makes sense. 

So if you want you can simply use the below code ;) 

```javascript
        if (sourceButton) {
            let conversationCached = false;
            window.addEventListener('TBNewPage', () => {
                conversationCached = false;
            });

            TB.listener.on('author', event => {
                if (event.detail.type !== 'TBmodmailCommentAuthor') {
                    return;
                }
                const $target = $(event.target);

                $target.closest('.Message__header').append('<button class="tb-source-button tb-general-button">source</button>');

                $('.tb-source-button').click(async e => {
                    // Something is causing the listener to be triggered multiple times.
                    e.stopImmediatePropagation();
                    const $currentSourceBtn = $(e.currentTarget.parentElement);

                    // Getting the ID of the message on which the button was clicked.
                    const activeMessageID = event.detail.data.comment.id;
                    const $currentSourceField = $(`#tb-source-${activeMessageID}`);

                    // Toggling the source
                    if ($currentSourceField.length) {
                        // If the source field exists, toggle it

                        $currentSourceField.toggle();
                    } else {
                        // If the source field is not present (has not been requested yet), request it and create
                        // a div+textarea with the source.

                        if (!$currentSourceBtn.closest('.Thread__message').has('.tb-source-field').length) {
                            let conversationInfo;
                            if (conversationCached) {
                                conversationInfo = await TBStorage.getCache('NewModmailPro', 'current-conversation');
                            } else {
                                // Fetch and store the conversation info in cache
                                const currentID = event.detail.data.post.id;
                                conversationInfo = await TBApi.apiOauthGET(`/api/mod/conversations/${currentID}`).then(r => r.json());
                                TBStorage.setCache('NewModmailPro', 'current-conversation', conversationInfo);
                                conversationCached = true;
                            }
                            // Getting the body in markdown from selected message
                            const messageSource = conversationInfo.messages[activeMessageID].bodyMarkdown;

                            $currentSourceBtn.closest('.Thread__message').append(`
                                <div class="tb-source-field" id="tb-source-${activeMessageID}">
                                    <textarea readonly></textarea>
                                </div>`);
                            $(`#tb-source-${activeMessageID} textarea`).text(messageSource);
                        }
                    }
                });
            });
        }
```

creesch (approved):

PR #343, opened 2020-07-28
Generate url_* tokens if not present in API info

### LINKED ISSUE #315: `uri_*` removal reason tokens are broken
Apparently the `/api/info.json` route used to return `uri_title` and `uri_body` properties, which were pre-URI-encoded versions of the title and body of the specified post/comment, and [we exposed those properties as tokens](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/modules/removalreasons.js#L232-L233) for use in embedding the title and body in prefilled links (e.g. [this use case pointed out in discord](https://discordapp.com/channels/535490452066009090/535490561478492161/712381420789563443)). The API has since removed those properties which means the tokens have stopped working.

I think it would make sense to manually fill these in by URL encoding the `title` and `body` fields we already get, rather than relying on the API.

Geo1088:
Fixes #315. Committed on github, test before merging.

### DIFF
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -229,8 +229,8 @@ function removalreasons () {
                     domain: info.domain,
                     body: info.body,
                     raw_body: info.raw_body,
-                    uri_body: info.uri_body,
-                    uri_title: info.uri_title,
+                    uri_body: info.uri_body || encodeURIComponent(info.uri_body),
+                    uri_title: info.uri_title || encodeURIComponent(info.title),
                     reasons: [],
                     get reasons_posts () {
                         return this.reasons.filter(r => r.removePosts || r.removePosts === undefined);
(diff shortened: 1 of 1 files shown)


extension/data/modules/removalreasons.js
```diff
@@ -229,8 +229,8 @@ function removalreasons () {
                     domain: info.domain,
                     body: info.body,
                     raw_body: info.raw_body,
-                    uri_body: info.uri_body,
-                    uri_title: info.uri_title,
+                    uri_body: info.uri_body || encodeURIComponent(info.uri_body),
```
creesch, line 232: I think you made a copy paste error here as you use the same variable that would already have failed. Probably want `info.body` ?

Geo1088: oof

```suggestion
                    uri_body: info.uri_body || encodeURIComponent(info.body),
```

Issue #198, opened 2019-11-21
Convert to ES6 module syntax


eritbh:
All our platforms support the module specification from ES6, and also support the currently stage 4 proposal for dynamic `import()`. These are all we need to convert the existing module structore to *actual* modules, allowing us to greatly simplify the init process.

Rather than waiting for an event and registering themselves, each module can export a TBModule object which will be imported and registered elsewhere. This will also allow us to avoid IIFEs wrapping everything, because module code doesn't expose declarations to other files unless they are explicitly `export`ed.

We can't make content scripts act as modules straight away, but we can use an async function to `import()` module code. Reference: https://stackoverflow.com/questions/48104433/how-to-import-es6-modules-in-content-script-for-chrome-extension

Geo1088:
I think this issue should also cover a restructuring of how we define modules. A couple thoughts:

- We should separate "always enabled" modules like config, oldreddit, etc from the user-configurable modules in some way. Currently, "module" is used in a couple different senses, meaning both "a part of Toolbox with its own options that users can disable", or "a chunk of related code", or both, depending on context. I think it makes sense that if the user can't enable or disable something, it doesn't need all the same overhead that modules with configurable settings, etc. have.

- Maybe this is just me, but `TB.Module` has a kinda strange interface at the moment, requiring you to create an instance and then patch properties manually in order to set configuration and defaults for the module. I'd prefer to revise this API so that all the necessary configuration is kept as an object parameter inside the constructor, e.g.
  ```js
  const self = TB.Module({
  	shortname: 'something',
  	enabledByDefault: true,
  	beta: false,
  	dev: false,
  	settings: {
  		// etc...
  	},
  }, async function init () {
    // init code can go here
  });
  ```

- The current method of getting settings values on init is clunky. It would probably be easier to have `TB.Module` instances get passed a settings object on init for convenience, which would probably also require some changes to the storage code.

- `TB.Module` right now has a prototype get/set pair for `shortname` that autofills the shortname if one isn't passed explicitly. We should trash this and ensure that all modules have explicit shortnames.

- Are those module profiling methods used anywhere? If not, we should probably get rid of them too. I'd rather see a separate file for debug/perf functions like that if they do need to be used.

[...]

PR #352, opened 2020-08-10
Add no-shadow to eslintrc

Geo1088:
Apparently `no-shadow` isn't part of `eslint:recommended`. It should be, but that's not up to me.

This adds `no-shadow` to our config. This rule reports an error whenever a variable name from an upper scope is reused in a lower scope. This prevents uncertainty in which identically-named variable is actually being referenced in code. `builtinGlobals` ensures that we aren't shadowing the names of built-in objects, and `hoist: all` makes the following invalid:

```js
if (someCondition) {
	const foo = 1;
	// errors because `foo` is declared in the upper scope, even though it comes after this block
}

const foo = 2;
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,6 +32,13 @@
             "error",
             "global"
         ],
+        "no-shadow": [
+            "error",
+            {
+                "builtinGlobals": true,
+                "hoist": "all"
+            }
+        ],
         "no-prototype-builtins": "warn",
         "no-console": 0,
         "no-multi-spaces": "error",


Geo1088:
hm, no kidding we didn't have this enabled before, travis reports almost 200 errors caused by this change. Marking as WIP until I get around to fixing those issues.

Geo1088:
abandoning this, something about our eslint environment is wonky and marking lots of things as shadowed when they clearly aren't.

image

Probably not worth revisiting this until after #198, because that will allow us to clean up our mess of global variables somewhat. So this is a long way off from being practical.

PR #357, opened 2020-09-01
feat(Modbar): add tooltip to subreddit names

SpyTec:
In moderated subreddits the subreddit names does not have a tooltip
that displays the name. This can be an issue for subreddits with a very
long name


creesch:
This doesn't solve the visual issue does it? 

image


Would be nice to include a fix for that as well in this PR.

creesch:
NEVERMIND, just saw the other pr.

PR #362, opened 2020-09-05
fix(Modmacro): every context to true when undefined

SpyTec:
If a context in mod macro is undefined, we default to checked


creesch:
lgtm

PR #367, opened 2020-09-07
fix(RR): Reverts textarea set from placeholder attribute

SpyTec:
With newest update, placeholder is being used as an actual value instead of a
placeholder. This is unconventional and for most people, a breaking
change for many.

### DIFF
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -695,9 +695,7 @@ function removalreasons () {
                 // Get input from HTML-formatted reason
                 const htmlReason = $this.find('.reason-content');
                 htmlReason.find('select, input, textarea').each(function () {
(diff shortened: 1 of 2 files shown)


Geo1088 (approved):
obligatory @molenzwiebel was right, I was wrong - still not entirely sure what I was thinking when I requested this change in #305. LGTM, labeling as a bug since creesch had reports of breaking existing configs (though I'm not sure how the placeholder was being used previously since it wasn't on the attribute whitelist).

creesch:
>  (though I'm not sure how the placeholder was being used previously since it wasn't on the attribute whitelist).

People probably put it in forgetting about it, now it is surfacing where they aren't expecting it. In those cases it is better to simply show it as a placeholder instead of an actual value.

PR #368, opened 2020-09-09
Misc. Deepscan fixes

Geo1088:
Fixes the four outstanding issues we have in Deepscan, including one that results in flyingsnoo not animating correctly when you pop the balloons.


SpyTec:
LGTM, is this all for this PR?

PR #380, opened 2020-10-03
Unify window style

### LINKED ISSUE #240: Unify markup/classes for dialogs/windows/notifications

Geo1088:
This PR unifies the `.tb-window-*`, `.tb-popup-*`, and `.tb-notification-*` classes under a single `.tb-window-*` scheme including modifier classes. Also adds an argument to `TBui.overlay` to control the tab bar orientation, to make the profile overlay less of a special case, and adds documentation to the arguments of that function. The overlay CSS was also modified to not rely on containing a specific type of window in case that's desirable in the future (though `TBui.overlay` still generates large static windows and nothing else).

- `TBui.popup()` returns a `.tb-window.tb-window-draggable`
- `TBui.overlay()` uses a `.tb-window.tb-window-large.tb-window-vertical-tabs`
- `TBui.showNotification()` uses a basic `.tb-window`

Technically, there are still slight differences in the markup generated by `TBui.popup` and `TBui.overlay`, but those differences are mostly in how data for multiple tabs is stored. This PR is only concerned with unifying the classes controlling visual appearance; a future PR can handle unifying the internal structure, since that will probably require breaking out a dedicated tabber component and retrofitting it into the existing UI generators.

Supercedes #327. Fixes #240.

### DIFF
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -490,8 +490,8 @@ function modbar () {
             if ($this.data('module')) {
                 $tb_help_mains.data('module', $this.data('module'));
             }
-            $('.tb-personal-settings .tb-window-wrapper .tb-window-tab').hide();
-            $(`.tb-personal-settings .tb-window-wrapper .tb-window-tab.${module}`).show();
+            $('.tb-personal-settings .tb-window .tb-window-tab').hide();
+            $(`.tb-personal-settings .tb-window .tb-window-tab.${module}`).show();
         }
 
         function checkHash () {
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -286,8 +286,8 @@ function modbutton () {
             // Hide the flair tab and message tab
             // TODO: add a "disabled" state, with tooltip, and use that instead
             // We can only edit flair in the current sub.
-                $popup.find('.tb-popup-tabs .user_flair').remove();
-                $popup.find('.tb-popup-tabs .send_message').remove();
+                $popup.find('.tb-window-tabs .user_flair').remove();
+                $popup.find('.tb-window-tabs .send_message').remove();
             }
 
             // get pre-definded ban message/note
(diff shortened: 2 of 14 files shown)


SpyTec (commented):
Went over all the changes, it seems to make sense. I still have to visually look it over by checking out the PR

creesch (commented):
Oddly enough personal notes doesn't seem to want to show on a regular frontpage (it is there hiding at the bottom) but does show in queue. Not sure what is going on there but it is specific to this PR. 

Other than that it looks good to me.

Geo1088:
Latest commit fixes that issue for me.

Issue #376, opened 2020-10-02
Modbutton ban message can become too long when expanding tokens


Geo1088:
When using the mod button to issue a ban, using repeated tokens in the message (e.g. `{body}{body}{body}`) can cause the final message's length to exceed the 1000 character cap, circumventing the `maxlength` on the text field and resulting in the ban not going through. We should expand tokens, check the message length, and if it exceeds the character limit, display the full message to let the user edit it down within the character limit before sending the ban.

Similar bugs probably exist in other places where tokens can be used as well. Might be worth making the solution to this a TBui routine for reusability.

SpyTec:
We already have a substring for message in TBApi though, it shouldn't cause any issues with the request

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/7a67e74379f4477fb6e83f597ae56fdbe3f6ac67/extension/data/tbapi.js#L390-L391

Geo1088:
huh... that's very strange, I was troubleshooting an issue on Discord the other day that the user said was solved by making the reason shorter... I'll have to look into this further.

PR #383, opened 2020-10-04
feat(Modbutton): Display char limit error on reason and duration

SpyTec:
Adds visual errors for ban reason and ban duration.

Related to #376

### DIFF
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const MAX_BAN_REASON_LENGTH = 300;
+const MAX_BAN_MESSAGE_LENGTH = 1000;
+
 function modbutton () {
     const self = new TB.Module('Mod Button');
     self.shortname = 'ModButton';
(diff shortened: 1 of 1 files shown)


extension/data/modules/modbutton.js
```diff
@@ -1,5 +1,8 @@
 'use strict';
 
+const MAX_BAN_REASON_LENGTH = 300;
+const MAX_BAN_MESSAGE_LENGTH = 1000;
```
Geo1088, line 4: I almost wonder if we should make these TBApi constants and use them when actually trimming the strings before sending too...

Geo1088 (approved):
LGTM 👍

extension/data/modules/modbutton.js
SpyTec: We most likely could. Though in TBApi we use substring with is index based instead of length based, so might not look as nice :P

Geo1088: Could change it to use `substr` instead, which is length-based :P

Totally extra credit though, I don't think it'll be a problem the way it is. Feel free to merge if you don't feel like doing that.

SpyTec: Way too lazy for that

PR #384, opened 2020-10-04
feat(Modmail): User-friendly permalink

### LINKED ISSUE #371: figure out how to use `legacy_first_message_id` in new modmail
For modmacros and ban macros we probably want to use it to fill in the `permalink` token (well we scrape it from the page but that is besides the point). 

For usernotes I think we should keep using the urls to new modmail itself as those are only used internally by mods. We probably want to create a new token along the lines of `newmodmail_permalink` for that purpose and [replace this line](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/modules/usernotes.js#L483).

Did I miss anything?

SpyTec:
Fixes #371

This takes the URL directly from HTML

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -967,7 +967,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 user,
                 author: user,
                 permalink,
-                permalink_newmodmail: permalink_newmodmail ? permalink_newmodmail : permalink,
+                permalink_newmodmail: permalink_newmodmail || permalink,
                 url: permalink,
                 domain,
                 id,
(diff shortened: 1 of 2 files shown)


extension/data/tbcore.js
```diff
@@ -962,6 +967,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 user,
                 author: user,
                 permalink,
+                permalink_newmodmail: permalink_newmodmail ? permalink_newmodmail : permalink,
```
Geo1088, line 970:
```suggestion
                permalink_newmodmail: permalink_newmodmail || permalink,
```

Geo1088 (changes requested):
One code style nit, otherwise LGTM.

Geo1088 (approved):

PR #386, opened 2020-10-05
Remove dependence on jQuery effects module

### LINKED ISSUE #105: Look at replacing jquery effects module calls.
https://api.jquery.com/category/effects/

Doing this and resolving #45 means we can switch to jquery slim.

eritbh:
Fixes #105. Lets us switch to jQuery's slim build. This is gonna take a while.

### DIFF
--- a/extension/data/modules/historybutton.js
+++ b/extension/data/modules/historybutton.js
@@ -315,7 +315,7 @@ function historybutton () {
               $markdown = $contentBox.find('.submission-markdown-text');
 
         if ($markdown.length > 0) {
-            $markdown.html(markdownReport).toggle();
+            $markdown.remove();
         } else {
             $contentBox.find('.table.domain-table').before(`<div class="submission-markdown">
                 <textarea class="tb-input submission-markdown-text">${markdownReport}</textarea>
--- a/extension/data/modules/modmatrix.js
+++ b/extension/data/modules/modmatrix.js
@@ -294,8 +294,13 @@ function modmatrix () {
         // Show percentages
         modMatrixSettings.find('table').append('<tr><td><label for="showpercentages">show percentages:</label></td><td><input id="showpercentages" type="checkbox" checked="checked" /></td></tr>');
         $('#showpercentages').change(function () {
-            $('#mod-matrix .action-percentage').toggle(this.checked);
-            $('#highlightpercentages').parent().parent().toggle(this.checked);
+            if (this.checked) {
+                $('#mod-matrix .action-percentage').css('display', '');
+                $('#highlightpercentages').parent().parent().css('display', '');
+            } else {
+                $('#mod-matrix .action-percentage').css('display', 'none');
+                $('#highlightpercentages').parent().parent().css('display', 'none');
+            }
         });
 
         // Highlight percentages
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -1224,7 +1224,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 count -= 1;
                 if (count <= 0) {
                     $body.find('#ratelimit-counter').empty();
-                    $body.find('#ratelimit-counter').hide();
+                    $body.find('#ratelimit-counter').css('display', '');
                     return count;
                 }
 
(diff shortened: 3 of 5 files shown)


extension/data/modules/modmatrix.js
```diff
@@ -294,8 +294,13 @@ function modmatrix () {
...
         $('#showpercentages').change(function () {
-            $('#mod-matrix .action-percentage').toggle(this.checked);
-            $('#highlightpercentages').parent().parent().toggle(this.checked);
+            if (this.checked) {
+                $('#mod-matrix .action-percentage').css('display', '');
+                $('#highlightpercentages').parent().parent().css('display', '');
+            } else {
+                $('#mod-matrix .action-percentage').css('display', 'none');
+                $('#highlightpercentages').parent().parent().css('display', 'none');
+            }
```
SpyTec, line 303:
```suggestion
            const displayToggle = this.checked ? 'none' : '';
            $('#mod-matrix .action-percentage').css('display', '');
            $('#highlightpercentages').parent().parent().css('display', '');
```

extension/data/tbcore.js
```diff
@@ -1224,7 +1224,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                 count -= 1;
                 if (count <= 0) {
                     $body.find('#ratelimit-counter').empty();
-                    $body.find('#ratelimit-counter').hide();
+                    $body.find('#ratelimit-counter').css('display', '');
```
SpyTec, line 1227: We're hiding here, shouldn't this be `display: none`?

extension/data/modules/historybutton.js
```diff
@@ -315,7 +315,7 @@ function historybutton () {
               $markdown = $contentBox.find('.submission-markdown-text');
 
         if ($markdown.length > 0) {
-            $markdown.html(markdownReport).toggle();
+            $markdown.remove();
```
creesch, line 318: This doesn't seem like correct behavior to me, as it means that with every toggle the html needs to be redone.

extension/data/tbcore.js
Geo1088: Stylesheet has `display: none` already, so `block` is needed to show it and having nothing in the inline style hides it.

extension/data/modules/modmatrix.js
Geo1088: Assuming you meant `.css('display', displayToggle);` for this suggestion? Makes sense though, I'll change it.

extension/data/modules/historybutton.js
Geo1088: The actual report text isn't regenerated, just the element it's displayed in. Given how much simpler this solution is and how infrequently anyone is ever going to toggle this button twice, the minor performance difference seemed worth it to me.

creesch: > how infrequently anyone is ever going to toggle this button twice

I wouldn't count on that considering the sort of odd behavior we have seen from people over the years :P

extension/data/tbcore.js
SpyTec: Gotcha, add a comment about it. By itself it is confusing

extension/data/modules/modmatrix.js
SpyTec: Yes, fixed 😂

extension/data/modules/historybutton.js
Geo1088: Reusing the same DOM elements helps for operations where we'd otherwise need to create and display large numbers of them at once, but creating and inserting one new textarea when a user clicks a button isn't going to result in the same sort of performance hit we've experienced in the past (e.g. with the usernotes manager deadlocking script execution and creating "page unresponsive" errors).

I wasn't able to detect any performance disadvantage of this method subjectively when I tested it, at least, and removing the element rather than mucking around with inline CSS makes the code a bit easier to understand.

eritbh:
This has been sitting around a while with no work being done, and I don't think it's really a good goal. regardless, given how stale this branch is work on something like this would basically have to be restarted from the beginning anyway. I'm closing this.

PR #387, opened 2020-10-05
Optimize watchPushState on old Reddit

larsjohnsen:
The previous solution required animation frames to be scheduled as
long as the page was visible, preventing the CPU from idling.

This change adds a passive listener in the background which send a
message to the content script when the tab has a new URL. The
extension API is used as there's no native Web API which triggers an
event when the URL is modified by history#pushState.

This change reduces Toolbox's CPU usage on my laptop from 2-3% from 0%. It is unlikely that this change has any significant impact on the redesign.

Before:
image

After:
image

### DIFF
--- a/extension/data/background/handlers/url_changed.js
+++ b/extension/data/background/handlers/url_changed.js
@@ -1,0 +1,9 @@
+'use strict';
+
+// Notifies tabs when their URL changes.
+// This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
+browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+    if (changeInfo.url) {
+        browser.tabs.sendMessage(tab.id, {action: 'tb-url-changed'});
+    }
+});
--- a/extension/data/background/handlers/url_changed.js
+++ b/extension/data/background/handlers/url_changed.js
@@ -3,13 +3,13 @@
 // Notifies tabs when their URL changes.
 // This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
 function handleWebNavigation ({tabId, frameId}) {
-    browser.tabs.sendMessage(tabId, {action: 'tb-url-changed'})
+    browser.tabs.sendMessage(tabId, {action: 'tb-url-changed'}, {frameId})
         .catch(error => {
             // Receiving end errors are not really relevant to us and happen a lot for iframes and such where toolbox isn't active.
             if (error.message !== 'Could not establish connection. Receiving end does not exist.') {
                 console.warn('tb-url-changed: ', error.message, error);
             }
-        }, frameId);
+        });
 }
 
 const filter = {url: [{hostContains: 'reddit.com'}]};
--- a/extension/data/background/handlers/url_changed.js
+++ b/extension/data/background/handlers/url_changed.js
@@ -2,14 +2,16 @@
 
 // Notifies tabs when their URL changes.
 // This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
-browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
-    if (changeInfo.url) {
-        browser.tabs.sendMessage(tab.id, {action: 'tb-url-changed'})
-            .catch(error => {
-                // Receiving end errors are not really relevant to us and happen a lot for iframes and such where toolbox isn't active.
-                if (error.message !== 'Could not establish connection. Receiving end does not exist.') {
-                    console.warn('tb-url-changed: ', error.message, error);
-                }
-            });
-    }
-});
+function handleWebNavigation ({tabId, frameId}) {
+    browser.tabs.sendMessage(tabId, {action: 'tb-url-changed'})
+        .catch(error => {
+            // Receiving end errors are not really relevant to us and happen a lot for iframes and such where toolbox isn't active.
+            if (error.message !== 'Could not establish connection. Receiving end does not exist.') {
+                console.warn('tb-url-changed: ', error.message, error);
+            }
+        }, frameId);
+}
+
+const filter = {url: [{hostContains: 'reddit.com'}]};
+browser.webNavigation.onReferenceFragmentUpdated.addListener(handleWebNavigation, filter);
+browser.webNavigation.onHistoryStateUpdated.addListener(handleWebNavigation, filter);
--- a/extension/data/background/handlers/url_changed.js
+++ b/extension/data/background/handlers/url_changed.js
@@ -4,6 +4,12 @@
 // This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
 browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
     if (changeInfo.url) {
-        browser.tabs.sendMessage(tab.id, {action: 'tb-url-changed'});
+        browser.tabs.sendMessage(tab.id, {action: 'tb-url-changed'})
+            .catch(error => {
+                // Receiving end errors are not really relevant to us and happen a lot for iframes and such where toolbox isn't active.
+                if (error.message !== 'Could not establish connection. Receiving end does not exist.') {
+                    console.warn('tb-url-changed: ', error.message, error);
+                }
+            });
     }
 });
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -1580,13 +1580,12 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
         const userProfile = /^\/user\/([^/]*?)\/?(overview|submitted|posts|comments|saved|upvoted|downvoted|hidden|gilded)?\/?$/;
         const userModMessage = /^\/message\/([^/]*?)\/([^/]*?)?\/?$/;
 
-        // This function after being first called will watch for pushstate changes.
         // Once a change is detected it will abstract all the context information from url, update TBCore variables and emit all information in an event.
         // NOTE: this function is a work in progress, page types are added once needed. Currently supported pages where context are provided are:
         // NewModmail: listings, conversations, create
         // reddit frontpage: sorting
         // subreddits: listing including sorting, submissions, submissions with permalink
-        function watchPushState () {
+        function refreshUrlContext () {
             const samePage = locationHref === location.href;
             if (!samePage) {
                 const oldHref = locationHref;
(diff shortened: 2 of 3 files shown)


extension/data/tbcore.js
```diff
@@ -1580,13 +1580,12 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
...
-        // This function after being first called will watch for pushstate changes.
         // Once a change is detected it will abstract all the context information from url, update TBCore variables and emit all information in an event.
         // NOTE: this function is a work in progress, page types are added once needed. Currently supported pages where context are provided are:
         // NewModmail: listings, conversations, create
         // reddit frontpage: sorting
         // subreddits: listing including sorting, submissions, submissions with permalink
-        function watchPushState () {
+        function refreshUrlContext () {
             const samePage = locationHref === location.href;
             if (!samePage) {
```
Geo1088, line 1590: Since the background page is now responsible for checking if the URL has changed, not the consumer, we should remove this check and the `locationHref` variable.

extension/data/background/handlers/url_changed.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+// Notifies tabs when their URL changes.
+// This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
+browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+    if (changeInfo.url) {
+        browser.tabs.sendMessage(tab.id, {action: 'tb-url-changed'});
```
Geo1088, line 7: I wonder if it would be a good idea to send `changeInfo.url` along with this message and have `refreshUrlContext` take the new URL as an argument rather than relying on `location.href`. I doubt this event would ever be emitted before `location.href` is updated or anything like that, but it feels more intuitive to me to include the event data with the event itself.

Just thinking out loud here, not a required change.

extension/data/tbcore.js
```diff
@@ -1714,10 +1713,11 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
                     window.dispatchEvent(new CustomEvent('TBNewPage', {detail: contextObject}));
                 }, 500);
             }
-            requestAnimationFrame(watchPushState);
         }
 
-        watchPushState();
+        refreshUrlContext();
```
Geo1088, line 1718: Any chance we can catch the `tabs.onUpdate` event emitted when the page is initially loaded and avoid having to call this manually the first time? Then we could just inline `refreshUrlContext` into the `window.addEventListener` call.

Geo1088 (changes requested):
Very cool, I never considered using the `browser.tabs` API for this! I have a couple thoughts about cleaning up the surrounding code to account for this change, but this looks solid otherwise.

extension/data/tbcore.js
creesch: Nope, don't think so otherwise it wouldn't work on page load.

extension/data/background/handlers/url_changed.js
Geo1088: Oh, also, we should catch any errors thrown by `sendMessage` as otherwise it'll spam the console with "No receiving end" errors when trying to send these updates to non-Reddit tabs where Toolbox isn't listening for the event.

extension/data/tbcore.js
creesch: I was just thinking about this and I am not sure that is a good idea as it is a recipe for race conditions where the state is being send to a frontend that hasn't loaded in at the time.

extension/data/background/handlers/url_changed.js
larsjohnsen: Fixed the `sendMessage` issue.

I chose not to add `changeUrl.url` to the message since:
- A tab may have multiple URLs if they have embedded frames
- If the URL is changed multiple times in short succession, the message may contain an outdated `href`.

extension/data/tbcore.js
larsjohnsen: Agreed with @creesh.

extension/data/background/handlers/url_changed.js
larsjohnsen: > A tab may have multiple URLs if they have embedded frames

On a second look, this might not be valid. Editing the URL of embedded frames does not trigger the `onUpdated` callback. This may also restrict the usefulness of this PR, as `refreshUrlContext` will only be triggered if Toolbox is running as the top frame.

creesch: Hrm... That might not be that much of an issue as we only embed toolbox on old reddit for some things. In which case it already gets triggered on page load.

creesch: We probably should make sure though. As I was looking into this I ran into `browser.webNavigation`, which might be an alternative to look into. [MDN info](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation)

larsjohnsen: I tested it, and `webNavigation` seems to fit the use case even better! It does however require that a new permission is declared, but I am uncertain whether the user needs to grant additional permissions as it just requires access to "Read your browsing history", which is already required by the "tabs" permission. https://developer.chrome.com/extensions/permission_warnings#webNavigation

extension/data/tbcore.js
Geo1088: That's what I figured too, was just crossing my fingers it wouldn't be the case :P

extension/data/background/handlers/url_changed.js
Geo1088: Hopefully Firefox's permission system is similar so we can avoid extra prompts, but honestly I think it's worth it to add the new permission and handle this more completely either way. We don't rely on being able to navigate Reddit in frames right now, but there may be cases where that's useful in the future.

Whether or not the new URL is actually passed to the callback or not isn't particularly important to me, since it's implied from these docs that `location` in the relevant frame is updated before any navigation events are sent. Creesch gets final say on whether a new manifest permission for `webNavigation` is worth it.

extension/data/tbcore.js
Geo1088: @creesch and I talked about this point briefly on Discord and basically concluded we didn't quite know what each other were talking about, so I'll try to clarify what I mean here. @larsjohnsen let me know if this makes sense.

Previously, we detected URL changes by keeping track of the previous frame's `location.href` in the variable `locationHref` and comparing it to the current frame's `location.href`, storing the result of that comparison in `samePage`. Then, if `samePage` is false, we do the logic for the URL changing. In this PR, detecting URL changes is handled by the background page, which means we don't need `samePage`, and since the inner logic doesn't care about the previous URL either, we don't need `locationHref`.

creesch: Now that I have thought about it and seeing your reasoning here I am not sure if we really should remove it. (depending on what API we end up using). As Lars pointed out elsewhere there can be multiple events fired from the background page for whatever reason. We really only want to trigger if there is actually a page change.

Geo1088: Would using the `webNavigation` API rather than `tabs` still run the risk of emitting multiple changes? It doesn't seem like it to me. I'm inclined to trust the API telling us when the page changes and not double-check things for ourselves if they're already guaranteed, but if `webNavigation` doesn't help (or if you decide we don't want to introduce a new permission to use it), then I can understand this.

extension/data/background/handlers/url_changed.js
creesch: Did a bit of testing based on the below code and `onHistoryStateUpdated` seems to do the trick rather well including frames. It doesn't trigger on hash changes to the url for which `onReferenceFragmentUpdated` would be needed but that isn't relevant for this PR. It does need the `webNavigation` permission to be able to function though. 

```javascript

const filter = {
    url:
    [
        {hostContains: 'reddit.com'},
    ]
};

function webNavigationHandling (details) {
    console.log(details)
    console.log(`url: ${details.url}`);
    console.log(`Transition type: ${details.transitionType}`);
    console.log(`Transition qualifiers: ${details.transitionQualifiers}`);
}

browser.webNavigation.onHistoryStateUpdated.addListener(webNavigationHandling, filter);

browser.webNavigation.onReferenceFragmentUpdated.addListener(webNavigationHandling, filter);
```

extension/data/tbcore.js
creesch: Btw, we still need TBnewPage to fire on page load as well, which is what my original point was about. Though on page load we can probably just fire the function itself without a comparison.

extension/data/background/handlers/url_changed.js
Geo1088: The reference fragment bit doesn't seem super important to me at this point, because according to my testing Reddit refuses to let you click fragment links in comments without refreshing the page anyway. Unless we link to settings that way internally, it should only get triggered when navigating on a wiki page's table of contents or something like that. Can't hurt to be more complete though.

extension/data/tbcore.js
Geo1088: I was assuming that would be the case. We currently handle page load by calling the function manually, and I don't think this PR changes that. That's also what [this conversation](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/387#discussion_r499892017) was about.

I'm gonna resolve this review comment now though - I agree that it's fine to leave this as-is.

extension/data/background/handlers/url_changed.js
```diff
@@ -0,0 +1,17 @@
...
+// Notifies tabs when their URL changes.
+// This is triggered in the background as there's no native event which is emitted when `history#pushState` modifies the URL.
+function handleWebNavigation ({tabId, frameId}) {
+    browser.tabs.sendMessage(tabId, {action: 'tb-url-changed'})
+        .catch(error => {
+            // Receiving end errors are not really relevant to us and happen a lot for iframes and such where toolbox isn't active.
+            if (error.message !== 'Could not establish connection. Receiving end does not exist.') {
+                console.warn('tb-url-changed: ', error.message, error);
+            }
+        }, frameId);
```
Geo1088, line 12: I don't think `.catch()` takes a second argument, was this `frameId` left in by mistake? Maybe you meant to make it part of the log message or something?

Geo1088 (changes requested):
LGTM, tested locally and everything works as expected. I'm happy with the state this is in now after our other conversations, just one more nitpicky thing I noticed on your latest push. I'll wait for @creesch's approval before merging as well.

PR #394, opened 2020-10-10
Custom pages for syntax highlighter

### LINKED ISSUE #92: Custom pages for syntax highlighter
It would be awesome if there was an option to list additional wiki paths that the syntax highlighter should work on. I'm planning to put some script/bot configuration in there, and being able to make certain pages be opened with the syntax highlighting editor would be super neat.

For bonus points, ability to select type (yaml, json, css, etc) for each path, and maybe having wildcard paths.

itskieranp:
Resolves #92 

I was looking to see if this was possible myself, and found issue #92 via a [recent Reddit thread](https://www.reddit.com/r/toolbox/comments/go7cas/is_it_possible_to_add_other_wiki_pages_to_syntax/). Felt I was confident enough to take a look at the code, and as it doesn't appear that anybody is working on this, I gave it a go.

This allows the user to specify which wiki pages (besides the stylesheet as it's currently referenced with `/about/stylesheet` but not `/wiki/config/stylesheet` - let me know if you want that to be added as a checkbox toggle for consistency's sake and I'll probably handle both on the same checkbox), and the defaults are the current list of pages Toolbox supports: `config/automoderator`, `automoderator-schedule`, and `toolbox`, which can be removed if the user wishes to.

To account for people putting "json" instead of "javascript" as the language, that gets changed on-the-fly as the page loads, hopefully pre-empting the inevitable support requests. (And given I did that myself for a while until I realised the CodeMirror mode was called `javascript` not `json` - I know I'd be one of them)

Feel free to get in touch - I've joined the Discord and my PFP is the same as on here.

### DIFF
--- a/extension/data/modules/syntax.js
+++ b/extension/data/modules/syntax.js
@@ -17,10 +17,10 @@ function syntax () {
         default: {
             'config/automoderator': 'yaml',
             'automoderator-schedule': 'yaml',
-            'toolbox': 'javascript',
+            'toolbox': 'json',
         },
-        labels: ['page', 'language'], // language is one of [css,javascript,markdown,yaml] - more can be added to libs/codemirror/mode - will detect "json" and convert to "javascript"
-        title: 'In addition to the CSS, the following wiki pages get the specified code formatting. Language is one of css, javascript, markdown, or yaml',
+        labels: ['page', 'language'], // language is one of [css,json,markdown,yaml] - more can be added to libs/codemirror/mode - will detect "json" and convert to "javascript"
+        title: 'In addition to the CSS, the following wiki pages get the specified code formatting. Language is one of css, json, markdown, or yaml',
     });
     self.register_setting('selectedTheme', {
         type: 'syntaxTheme',
@@ -184,21 +184,6 @@ function syntax () {
             });
         }
 
-        // Are we on a wiki page from the list in the settings?
-        let onWikiPage = false;
-        for (const page in wikiPages) {
-            if (Object.prototype.hasOwnProperty.call(wikiPages, page)) { // ESLint guard-for-in
-                const pagePathRegex = new RegExp(`/wiki/(edit|create)/?${page}/?$`);
-                if (location.pathname.match(pagePathRegex)) {
-                    onWikiPage = page;
-                    // javascript mode used for JSON files, edit name to match.
-                    if (wikiPages[onWikiPage].toLowerCase() === 'json') {
-                        wikiPages[onWikiPage] = 'javascript';
-                    }
-                    break; // no need to keep checking once we've found we're on a listed page
-                }
-            }
-        }
         // Are we on a wiki edit or create page?
         const wikiRegex = /\/wiki\/(edit|create)/;
         if (location.pathname.match(wikiRegex)) {
@@ -209,15 +194,29 @@ function syntax () {
                     if (location.pathname.match(pagePathRegex)) {
                         // we've checked the current page is the edit page for one of the pages in the settings, replace the textarea with CodeMirror
 
-        // Here we deal with pages other than the stylesheet, from the list in the settings.
-        if (onWikiPage) {
-            let defaultMode = 'default';
                         let miscEditor;
                         const $editform = $('#editform');
 
-            if (wikiPages[onWikiPage] !== '') {
-                defaultMode = wikiPages[onWikiPage];
-            }
+                        // let's get the type and convert it to the correct mimetype for codemirror
+                        let mimetype;
+                        switch (wikiPages[page].toLowerCase()) {
+                        case 'css':
+                            mimetype = 'text/css';
+                            break;
+                        case 'json':
+                            mimetype = 'application/json';
+                            break;
+                        case 'markdown':
+                        case 'md':
+                            mimetype = 'text/markdown';
+                            break;
+                        case 'yaml':
+                            mimetype = 'text/x-yaml';
+                            break;
+                        default:
+                            mimetype = 'text/markdown';
+                            wikiPages[page] = 'markdown';
+                        }
 
                         // Class added to apply some specific css.
                         $body.addClass('mod-syntax');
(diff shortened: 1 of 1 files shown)


extension/data/modules/syntax.js
```diff
@@ -173,19 +184,32 @@ function syntax () {
...
+                    if (wikiPages[onWikiPage].toLowerCase() === 'json') {
+                        wikiPages[onWikiPage] = 'javascript';
+                    }
+                    break; // no need to keep checking once we've found we're on a listed page
+                }
+            }
+        }
+
+        // Here we deal with pages other than the stylesheet, from the list in the settings.
+        if (onWikiPage) {
```
creesch, line 204: Why is this a second check and not simply included in the above bit?

creesch (changes requested):
> To account for people putting "json" instead of "javascript" as the language, that gets changed on-the-fly as the page loads, hopefully pre-empting the inevitable support requests. (And given I did that myself for a while until I realised the CodeMirror mode was called javascript not json - I know I'd be one of them)


I am fairly sure we should just put in json and not javascript. While the mode file for it is called 'javascript' it actually does some specific stuff for json, [see this](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/libs/codemirror/mode/javascript.js#L914) MIME type definition. 

It also looks like you refer to the language rather than the mime types like we did before. Additionally I am not sure how codemirror is going to handle faulty languages which currently you don't account for. 

The latter two can easily be accounted for by implementing a little switch statement when encountering a wiki page on line 195 doing something like this: 

```javascript
switch (wikiPages[onWikiPage]) {
  case 'css':
    wikiPages[onWikiPage] = 'text/css';
    break;
  case 'json':
    wikiPages[onWikiPage] = 'application/json';
    break;
  case 'markdown':
    wikiPages[onWikiPage] = 'text/markdown';
    break;
  case 'yaml':
    wikiPages[onWikiPage] = 'text/x-yaml';
    break;
  default:
    wikiPages[onWikiPage] = 'text/markdown';
}
```

extension/data/modules/syntax.js
```diff
@@ -173,19 +184,32 @@ function syntax () {
             });
         }
 
-        // Here we deal with automod and toolbox pages containing json.
-        if (location.pathname.match(/\/wiki\/(edit|create)\/(config\/)?automoderator(-schedule)?\/?$/)
-            || location.pathname.match(/\/wiki\/edit\/toolbox\/?$/)) {
+        // Are we on a wiki page from the list in the settings?
+        let onWikiPage = false;
+        for (const page in wikiPages) {
```
creesch, line 189: It is probably better to start by checking if we are on a wiki edit page to begin with.

itskieranp:
All the comments have been addressed. Thanks for the prompt feedback on this!

extension/data/modules/syntax.js
```diff
@@ -209,15 +194,29 @@ function syntax () {
...
+                        case 'markdown':
+                        case 'md':
+                            mimetype = 'text/markdown';
+                            break;
+                        case 'yaml':
+                            mimetype = 'text/x-yaml';
+                            break;
+                        default:
+                            mimetype = 'text/markdown';
+                            wikiPages[page] = 'markdown';
```
creesch, line 218: Is this still supposed to be in here?

itskieranp: The intention is to update the settings so that an invalid setting such as "dfugjhdkjghnk" gets changed to "markdown" in the UI, and I found when I was testing that modifying the settings object updates the settings that are displayed in the settings UI (and presumably stored in settings storage as well).

creesch (approved):

extension/data/modules/syntax.js
```diff
@@ -12,6 +12,16 @@ function syntax () {
         default: true,
         title: 'Enable word wrap in editor',
     });
+    self.register_setting('wikiPages', {
+        type: 'map',
+        default: {
+            'config/automoderator': 'yaml',
+            'automoderator-schedule': 'yaml',
+            'toolbox': 'json',
+        },
```
Geo1088, line 21: Should also contain `'config/stylesheet': 'css'` to CSS highlighting works on `/wiki/edit/config/styesheet` as well as `/about/stylesheet`

extension/data/modules/syntax.js
```diff
@@ -173,77 +184,101 @@ function syntax () {
...
-            || location.pathname.match(/\/wiki\/edit\/toolbox\/?$/)) {
-            let miscEditor;
-            const $editform = $('#editform');
-            let defaultMode = 'default';
+        // Are we on a wiki edit or create page?
+        const wikiRegex = /\/wiki\/(edit|create)/;
+        if (location.pathname.match(wikiRegex)) {
+            // Are we on a page from the list in the settings?
+            for (const page in wikiPages) {
+                if (Object.prototype.hasOwnProperty.call(wikiPages, page)) { // ESLint guard-for-in
```
Geo1088, line 192: Prefer `for...of` iterating over `Object.keys(obj)` or `Object.entries(obj)` instead of `for...in` with a guard. For example:

```js
for (const [page, language] of Object.entries(wikiPages)) {
	// ...
}
```

Or, even better, why not forgo the `for` loop altogether and get the name of the current page from the regexp at the beginning?

```js
// Are we on a wiki edit or create page?
const wikiMatch = location.pathname.match(/\/wiki\/(edit|create)\/([a-z0-9-_/]*[a-z0-9-_])/);
if (!wikiMatch) {
    return;
}

// Are we on a page from the list in the settings?
const page = wikiMatch[2];
if (!wikiPages[page]) {
    return;
}

// We now know the name of the current page and can start making the editor and looking up the mime type
```

Geo1088 (changes requested):
Good work so far! A couple thoughts from me.

extension/data/modules/syntax.js
creesch: @Geo1088 `/about/stylesheet` is handled else where and should not be included.

Geo1088: I'm aware, the page under `/about` is handled manually since it's not technically a wiki edit page but a separate interface. The stylesheet can be edited in the normal wiki edit interface too, though, which is what I'm saying should be handled.

itskieranp: I didn't add it originally so as not to cause confusion if someone removed it - if we add that as a default, should we then detect if it's there before running on /about/stylesheet?

creesch: @itskieranp I'd say only the wiki page for it not the `about` one.

Geo1088: Agreed, probably not worth it to add more custom logic there. Most people who don't want syntax highlighting on the stylesheet page will just turn the entire module off anyway.

itskieranp:
Alrighty, pushed some more commits addressing these requests.

Geo1088 (approved):
Nice, lgtm! Thanks for the PR~

Issue #388, opened 2020-10-06
Make profile overlay search accessible via URL options


Geo1088:
Suggested [here](https://www.reddit.com/r/toolbox/comments/j5rb80/suggestion_add_a_url_argument_to_search_by/). We probably wouldn't want to use a proper query string for this, but doing something similar to the hash fragments RES adds when scrolling through NER pages could probably be doable.

creesch:
Not sure how suitable our current code is for this but we already have some stuff to deal with hash parameters. 

We do have `TBHelpers.getHashParameter` https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbhelpers.js#L625-L641

Modbar uses it to open the settings dialogue https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/modules/modbar.js#L497-L530  (side note, that interval mechanism is a bit odd, we might at one point look at doing that in a different manner possibly hook it into the url change mechanism)

creesch:
Thinking about it, it would be great if we could work it into `TBNewPage` (or at least that area) and make it event based for modules to listen to. Something like `TBHashParams`.

creesch:
To continue my spam (I'll make this the last comment for now), it seems that `browser.tabs.onUpdated` does also trigger on hash changes which is good. So we can probably include something in the newpage trigger like this 

```javascript 
if (window.location.hash) {
        const hash = window.location.hash.substring(1);
        console.log(hash)

        // To make sure we only trigger on toolbox hashes we check that the first param starts with `tb`.
        // This because `tbsettings` is already used for settings. 
        if(hash.startsWith('?tb')) {
            const paramObject = {};
            const params = hash.split('&');
            params.forEach(param => {
                const keyval = param.split('=');
                const key = keyval[0].replace('?', ''),
                      val = keyval[1];

                paramObject[key] = val;
  
            });
            window.dispatchEvent(new CustomEvent('TBHashParams', {detail: contextObject}));
        }
}
```

creesch:
As the hashing mechanism is now implemented I figure I'd start work on this. Before I do however I want to properly work out the parameters used. So far I am thinking about the following 

- `tbprofile` *mandatory* which will be one of `overview`, `comments` or `posts`.  
- `user` *optional* if not given on when on a profile page it will use that user and otherwise give an error message. 
- `subreddit` *optional* when present will initiate a profile search on opening of profile.
[...]

PR #395, opened 2020-10-11
Move hash based actions to be event triggered

creesch:
Done in preparation for #388. This PR moves toolbox's current hash processing to an event based system which allows us to easily hook in other things as well.

Initially I also wanted to implement things so that when a relative url is clicked to the settings that the state of the url is changed instead it reloading the page. However there is some fuckery going on with reddit urls making that a bit more difficult than I'd like. I briefly debated simply doing it for all `a` elements that open toolbox settings (and just ignoring whatever url they are pointing to) but I have no idea how people link to settings in regards to mod onboarding so I decided against that. 

I'll do #388 in a separate PR once this is merged.

### DIFF
--- a/extension/data/tbcore.js
+++ b/extension/data/tbcore.js
@@ -1611,7 +1611,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
         // NewModmail: listings, conversations, create
         // reddit frontpage: sorting
         // subreddits: listing including sorting, submissions, submissions with permalink
-        function refreshUrlContext () {
+        function refreshPathContext () {
             const samePage = locationHref === location.href;
             if (!samePage) {
                 const oldHref = locationHref;
@@ -1741,10 +1741,10 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
             }
         }
 
-        refreshUrlContext();
+        refreshPathContext();
         refreshHashContext();
         window.addEventListener('tb-url-changed', () => {
-            refreshUrlContext();
+            refreshPathContext();
             refreshHashContext();
         });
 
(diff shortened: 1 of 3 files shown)


Geo1088:
> Initially I also wanted to implement things so that when a relative url is clicked to the settings that the state of the url is changed instead it reloading the page.

Almost wish there were a RES feature to revert Reddit's garbage link handling in *all* cases. I know it's a relatively new thing on old Reddit at least. There's a lot of on-the-fly `href` modification that happens in response to link clicks that ideally we'd be able to override, but I don't think it's worth it for us at this point.

Anyway... Before I dive into the code here, is there a reason fragments are being handled in a separate function from the existing URL change detection? My understanding is that the existing stuff should handle hash changes the same as any other part of the URL changing. Correct me if I'm wrong.

creesch:
> Anyway... Before I dive into the code here, is there a reason fragments are being handled in a separate function from the existing URL change detection?

Mostly because `refreshUrlContext()` already is fairly long and deals with actual page changes where this deals with hashes. So it made sense to me to separate the handling of the two to keep things readable. 

I started out with all the logic within `refreshUrlContext()` but that was a bit messy, so I separated the hash specific stuff to `refreshHashContext()` and called that from within `refreshUrlContext()` and at that point I decided I might as well separate them entirely so they can also be worked on independently more easily if need be.

extension/data/tbcore.js
```diff
@@ -1716,7 +1742,11 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
         }
 
         refreshUrlContext();
-        window.addEventListener('tb-url-changed', refreshUrlContext);
+        refreshHashContext();
+        window.addEventListener('tb-url-changed', () => {
+            refreshUrlContext();
```
Geo1088, line 1747: Maybe we should consider renaming `refreshUrlContext` to something like `refreshPathContext` since it's not responsible for the whole URL anymore?

Geo1088 (commented):
Maybe we should consider renaming `refreshUrlContext` to something like `refreshPathContext` since it's not responsible for the whole URL anymore? Otherwise LGTM.

PR #397, opened 2020-10-13
Make profile overlay search accessible via URL options fixes #388

### LINKED ISSUE #388: Make profile overlay search accessible via URL options

creesch:
This PR fixes #388 and makes use of the existing hash param syntax as is already used for settings. Available parameters are:"

- `tbprofile` *mandatory* which will be one of `overview`, `comments` or `posts`.  
- `user` *optional* if not given on when on a profile page it will use that user and otherwise give an error message. 
- `subreddit` *optional* when present will initiate a profile search on opening of profile.
- `content` *optional* when present will initiate a profile search on opening of profile.
- `sort` *optional* when present will open the profile with that sort method and if a search is initiated it will check `use sort selection`. 

Takes the `Always open toolbox profile overlay on reddit profiles.` setting in account as well to prevent conflicts there.

### DIFF
--- a/extension/data/modules/profile.js
+++ b/extension/data/modules/profile.js
@@ -880,7 +880,7 @@ function profilepro () {
         // Open profile bashed on hash params
         window.addEventListener('TBHashParams', event => {
             const listing = event.detail.tbprofile;
-            if (listing && listingTypes.includes(listing)) {
+            if (listingTypes.includes(listing)) {
                 let user;
                 // If we get a user from the params we go from there.
                 if (event.detail.user) {
--- a/extension/data/modules/profile.js
+++ b/extension/data/modules/profile.js
@@ -46,6 +46,7 @@ function profilepro () {
         let filterModThings = false;
         let hideModActions = false;
         let cancelSearch = true;
+        const listingTypes = ['overview', 'submitted', 'comments'];
 
         const alwaysTbProfile = self.setting('alwaysTbProfile'),
               directProfileToLegacy = self.setting('directProfileToLegacy'),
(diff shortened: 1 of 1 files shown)


extension/data/modules/profile.js
```diff
@@ -851,6 +876,47 @@ function profilepro () {
             }
             makeProfile(user, listing, options);
         });
+
+        // Open profile bashed on hash params
+        window.addEventListener('TBHashParams', event => {
+            const listing = event.detail.tbprofile;
+            if (listing && listingTypes.includes(listing)) {
```
Geo1088, line 883: Passing falsy values to `Array.includes()` is fine, so the `listing &&` doesn't seem necessary here
```suggestion
            if (listingTypes.includes(listing)) {
```

extension/data/modules/profile.js
```diff
@@ -851,6 +876,47 @@ function profilepro () {
...
+                    user = TBCore.pageDetails.pageDetails.user;
+
+                // Finally we simply return if we have no username to work with.
+                } else {
+                    TB.ui.textFeedback('No user present in parameters and not on profile page.', TBui.FEEDBACK_NEGATIVE);
+                    return;
+                }
+
+                const options = {
+                    sort: event.detail.sort || 'new',
```
Geo1088, line 901: `sort` is defaulted to `'new'` in the definition of `makeProfile`, doesn't need to be defaulted here too
```suggestion
                    sort: event.detail.sort,
```

Geo1088 (changes requested):
Makes sense to me, just a couple nits. Let's definitely make sure this gets documented.

extension/data/modules/profile.js
```diff
@@ -592,6 +593,9 @@ function profilepro () {
             const renew = options.renew || false;
             const after = options.after || '';
             const subreddit = options.subreddit || '';
+            const content = options.content || '';
+            const search = options.search || false;
+            const searchSort = options.searchSort || false;
```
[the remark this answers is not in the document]
creesch: Yup gonna leave that for future improvements but you are right there.

PR #408, opened 2020-10-24
Improve startup performance

creesch:
Still a draft but from initial testing it seems to make 1-2 seconds depending on circumstances. Effectively it moves the API calls to get user information forward so they can get started before the DOM is in an interactive state. I also moved most javascript around in the manifest so it should be loaded sooner. 

Still a bit of a draft as I put in a lot of console log statements with `performance.now()` timestamps to get insights in delays and I still need to investigate a few things. 

I also did play around with replacing all `$body` declarations with one `TBCore` variant with the idea that this might improve performance but that doesn't seem to any impact (maybe in memory usage though, didn't look into that). 

I also checked if disabling modules like `betterbuttons` which still scrape the page in an oldschool way had an effect but that also doesn't seem to be the case. 

There is still some more investigating to do to see if we can get out a bit more performance but all things considered from what I can see a lot of the perceived delays is mostly down to the sheer amount of things toolbox does with reddit. 

On a final note, on firefox performance seems slightly better compared to chromium based browsers both in visual aspect as in actual numbers.


extension/data/modules/oldreddit.js
```diff
@@ -267,16 +268,27 @@ function oldReddit () {
...
+                    timeoutArray.forEach(timeout => {
+                        setTimeout(() => {
+                            newModmailSidebar();
+                            newModmailConversationAuthors();
+                        }, timeout);
+                    });
                 }
             });
         }
+        profileResults('oldRedditInitDone', performance.now());
```
Geo1088, line 284: This profile is basically useless since all that's being done between here and `oldRedditInit`	is setting a bunch of timeouts.

extension/data/tbapi.js
```diff
@@ -7,6 +7,14 @@
 (function (TBApi) {
     const logger = TBLog('TBApi');
 
+    TBApi.modhash = '';
+    TBApi.NO_WIKI_PAGE = 'NO_WIKI_PAGE';
+    TBApi.WIKI_PAGE_UNKNOWN = 'WIKI_PAGE_UNKNOWN';
+
+    TBApi.setModHash = modhash => {
+        TBApi.modhash = modhash;
+    };
```
Geo1088, line 16: Er... why does this function exist...?

extension/data/tbcore.js
```diff
@@ -4,22 +4,22 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
     (function (TBCore) {
         // We need these before we can do anything.
         TBCore.userDetails = userDetails;
-        TBCore.modhash = userDetails.data.modhash;
+        TBCore.modhash = TBApi.modhash;
```
Geo1088, line 7: We should really only store this value in one place, why not just replace all `TBCore.modhash` with `TBApi.modhash`?

extension/data/tbcore.js
```diff
@@ -601,7 +599,7 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
             if (showClose) {
                 $noteDiv.append(`<i class="note-close tb-icons" title="Close">${TBui.icons.close}</i>`);
             }
-            $noteDiv.appendTo('body');
+            $noteDiv.appendTobody;
```
Geo1088, line 602:
```suggestion
            $noteDiv.appendTo($body);
```

extension/data/tbui.js
```diff
@@ -1,1952 +1,1963 @@
...
-        if (!TBui.getBestTextColor.cache[bgColor]) {
-            const textColors = ['black', 'white'];
-            TBui.getBestTextColor.cache[bgColor] = tinycolor.mostReadable(bgColor, textColors).toHexString();
-        }
-        return TBui.getBestTextColor.cache[bgColor];
-    };
-    TBui.getBestTextColor.cache = {};
-})(window.TBui = window.TBui || {});
+if (document.readyState === 'loading') {
+    document.addEventListener('readystatechange', TBuiInitWrapper, {
```
Geo1088, line 1958: Shouldn't this be restricted to only run for a certain ready state?

extension/data/tbui.js
```diff
@@ -1,1952 +1,1963 @@
 'use strict';
+function TBuiInitWrapper () {
```
Geo1088, line 2: You're introducing another nested function when it's not really necessary. The inner IIFE can be eliminated, and `TBuiInitWrapper` can take `TBui` as a parameter.

though, honestly, since this whole thing is only run once per page load, the whole IIFE being called with `window.TBui = window.TBui || {}` is kinda overkill. Could just as easily put a `const TBui = window.TBui = {}` at the top of the function and call it a day. That goes for all our files structured like this - not really a performance thing though

extension/manifest.json
```diff
@@ -134,45 +171,6 @@
                 "data/styles/codemirror/matchesonscrollbar.css",
                 "data/styles/codemirror/show-hint.css",
                 "data/styles/codemirror/themes.css"
```
Geo1088, line 173: Can we move styles up too? Shouldn't really make any performance difference, but does mean we can eliminate the `document_end` manifest section.

extension/data/tbstorage.js
```diff
@@ -229,47 +229,151 @@ function storagewrapper () {
...
+                    return;
+                }
+                console.log('sendInit6', performance.now());
+                // https://bugzilla.mozilla.org/show_bug.cgi?id=1380812#c7
+                // https://github.com/toolbox-team/reddit-moderator-toolbox/issues/98
+                if ((typeof InstallTrigger !== 'undefined' || 'MozBoxSizing' in document.body.style) && browser.extension.inIncognitoContext) {
+                    logger.error('firefox is in incognito mode, toolbox will not work.');
+                    return;
+                }
+                console.log('sendInit7', performance.now());
```
Geo1088, line 361: will probably want to remove these logs eventually

extension/data/tbcore.js
```diff
@@ -1857,71 +1852,11 @@ function initwrapper ({userDetails, newModSubs, cacheDetails}) {
 
 (function () {
     const logger = TBLog('TBCore init');
-    // wait for storage
-    function getModSubs (after, callback) {
```
Geo1088, line 1861: My brain is having a hard time deciphering the diff here. What's being gained by moving these functions to storage?

Geo1088 (commented):
Mostly makes sense to me so far. Left some thoughts.

extension/data/tbui.js
creesch: Nope, as long as it isn't loading it is fine.

extension/data/tbui.js
creesch: Maybe something to tackle later, I just followed how we are already doing it (Though that was probably also my doing :P)

extension/manifest.json
creesch: From what I could see that actually did slow down performance a bit or rather made intial loading seem to slow down.

Geo1088: Interesting. Not a big deal then.

extension/data/tbui.js
Geo1088: I'd really prefer we do something about it, at least in this specific instance, so we don't end up with a 3k line diff for this file... Github doesn't even render the diff by default unless you tell it to ignore whitespace changes.

extension/data/tbapi.js
creesch: To set the modhash :P Has to do with how toolbox is structured, when TBApi is loaded in the modhash isn't known yet. I wanted a clean way to set it once it is known.

Geo1088: Can we not just do `TBApi.modhash = stuff` from wherever the modhash is retreived?

extension/data/tbui.js
Geo1088: Hm, so it seems like it's always being called on `interactive` then. I'd rather call out the state we're targeting explicitly, but that might be too small a nitpick. I see how it works now.

extension/data/tbcore.js
creesch: Because we are fetching this data sooner before the DOM is out of a loading state effectively at the same time as we are setting up storage. Before this they resided in tbcore.  As this is also effectively stored data we need before initializing it made sense to me to move the functions here.

extension/data/tbcore.js
creesch: Hrm not sure why I did it like this to be honest...

extension/data/tbcore.js
creesch: Oh boy, leftover from me trying to replace all `$body` instances with a single one to see if that had any measurable effect on performance (it didn't..)

extension/data/tbapi.js
creesch: Possibly, but afaik we never directly update class/object values so this did seem like a bit cleaner. There also might have been a technical reason I convinced myself was there, but I can't remember now....

extension/data/tbui.js
creesch: Frankly seemed like a wasted check as it really doesn't matter as long as it isn't `loading`.

creesch:
Aside from the obvious stuff that still needs fixing before we merge this I do want to try to get some actual benchmarks with enough datapoints with a few scenarios to see actual benefits. Specifically I came across a article regarding performance from a chrome developer where it was pointed out that changing the DOM while it is still being rendered might actually hurt performance. 

With that in mind I still want to experiment with some specific scenarios and try to get tangible date points: 

- Baseline, master as it is now without the above scenarios. 
- Current PR. 
- Current PR with CSS moved to load earlier. 
- Current PR but a change where we don't do the complete init at `interactive` but `ready` 
- Same as above but with the CSS moved. 

I'd like to try to not only measure reported load times but also visible load times, at what point do we see toolbox being loaded. To that end I'll probably set up a simple test with selenium.

eritbh:
This branch hadn't been updated since before #518 (content script esm) so it was very outdated. I got bored and merged current `master` back into it; a lot of the original performance changes that were in this branch were tied directly to the old event-based module registration system, so they're not here anymore, but there are a couple changes still relevant:

- Using a single `$body` variable rather than recomputing `$('body')` in some spots
- Removing `TBCore.NO_WIKI_PAGE` etc which were just unnecessary re-exported duplicates of the same on `TBApi`

Diff is much smaller now, could probably be a pretty easy merge.

PR #414, opened 2020-11-16
Use ratelimit handler for OAuth requests

### LINKED ISSUE #248: Handle ratelimiting consistently

eritbh:
Fixes #248. Adds a ratelimit handler class to the background page which is used for OAuth requests. The handler supports parallel requests up to the size of the current ratelimit bucket, so this should theoretically improve performance for some mass actions (which were limited to 20 requests per batch). I'll try to do a performance comparison in a bit.

(Edit: I just noticed that some `old.reddit.com` requests are receiving ratelimit responses as well. I'll look into this further and adapt this.) ~~As only `oauth.reddit.com` responds with ratelimit headers~~, this PR converts the methods that were previously controlled by `forEachChunkedRateLimit` to use OAuth rather than modhash authentication. The `forEachChunkedRateLimit` function itself is replaced with a `Promise.all()` structure that queues all requests at once and waits for all requests to complete before calling the completion handler.

This is in a functional state, but does not yet include a "you are being ratelimited" notification.


creesch:
> (Edit: I just noticed that some old.reddit.com requests are receiving ratelimit responses as well. I'll look into this further and adapt this

Afaik all of them do or should in any case (maybe reddit did mess it up somewhere), otherwise we wouldn't have been able to have our current ratelimiting system in place.

Geo1088:
On further inspection it seems like they only send them for `.json` requests. Stuff like fetching unread messages returns ratelimits, but hitting `/api/remove` doesn't actually appear to in my testing. Which means I'm really not sure how the current system functions at all for comment nuking.

Requests against non-JSON routes do seem to impact the ratelimits, they just don't contain headers. Since I account for that in the ratelimiter code, I think we should be fine to just set up one ratelimiter for OAuth requests and another for non-OAuth. I'll do that today.

creesch:
Doesn't the current system proactively check for the ratelimit before doing bulk requests? 

Still odd that not all requests contain that information, I rather not move too much things to oauth as we have had some scattered reports of oauth not working for all people (specifically on firefox).

Geo1088:
Oh, right, that function sent a request to a wiki JSON page to get the ratelimit headers from there. That makes sense now.

If you'd rather not move bulk action processing to OAuth, I can revert those changes. It should be fine to let them use the existing ratelimiter without much trouble - it's designed in such a way that even if we don't get ratelimit information, it should handle itself smoothly, falling back to serial requests if it doesn't know how much ratelimit is remaining. I'll have to test whether old.reddit sends 429s if the limit is exceeded or if I need to do special retry handling there, but it shouldn't be a big deal.

creesch:
If reddit let's us know properly that we are ratelimited when it happens I have no issue with a new method centered around that. Has been a while since I experimented with it though so I don't know if reddit does actually do that properly or just picked another random error code and throws in some html in the response as an extra fuck you.

Geo1088:
This should be in a mergeable state now, but let's hold it until we're done with 5.5 bugfix updates.

PR #418, opened 2020-12-01
Added support for flair template IDs to the mod button and removal reasons

### LINKED ISSUE #296: Using post flair IDs alongside CSS classes and text
Currently, while using a removal reason, the only two options regarding flairs are to assign flair text and class. New Reddit introduced flair IDs.

Could toolbox assign flairs based on their IDs when applying a removal reason?

Xeoth:
You can now use flair templates when flairing a user or when removing a post with removal reasons. Closes #296

### DIFF
--- a/extension/data/tbapi.js
+++ b/extension/data/tbapi.js
@@ -339,12 +339,13 @@
      * @param {string} cssClass The flair's CSS class
      * @returns {Promise}
      */
-    TBApi.flairUser = (user, subreddit, text, cssClass) => TBApi.post('/api/flair', {
+    TBApi.flairUser = (user, subreddit, text, cssClass, templateID = '') => TBApi.post('/api/selectflair', {
         api_type: 'json',
         name: user,
         r: subreddit,
         text,
         css_class: cssClass,
+        flair_template_id: templateID,
         uh: TBCore.modhash,
     });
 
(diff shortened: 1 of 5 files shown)


extension/data/tbapi.js
```diff
@@ -338,12 +339,13 @@
      * @param {string} cssClass The flair's CSS class
      * @returns {Promise}
      */
-    TBApi.flairUser = (user, subreddit, text, cssClass) => TBApi.post('/api/flair', {
+    TBApi.flairUser = (user, subreddit, text, cssClass, templateID = '') => TBApi.post('/api/selectflair', {
```
Geo1088, line 342: `flairUser` has a default for `templateID` but `flairPost` doesn't. I don't think we need to default it to an empty string, do we? Letting it be `undefined` should be fine.
```suggestion
    TBApi.flairUser = (user, subreddit, text, cssClass, templateID) => TBApi.post('/api/selectflair', {
```

Geo1088 (changes requested):
Looks good, seems to work nicely! Left notes on a couple minor code quality things. Adding "needs API docs" label so I remember to add the new removal reasons JSON key to the API docs later.

extension/data/tbapi.js
Xeoth: I actually wasn't sure about that, so I put it in there just in case having it as `undefined` would cause any issues, but you're most probably right.

Geo1088 (approved):

PR #421, opened 2020-12-02
Fix ambiguous sentence meaning in history button

### LINKED ISSUE #419: Fixing ambiguous sentence meaning in History Button
Text in the History button looks like this:
```
Available history:
790 submissions
283 comments of those 264 (93%) are in their own posts (commented as OP).
```

If I am reading this correctly I believe the last line is missing a comma after **comments**?
```
283 comments, of those 264 (93%) are in their own posts (commented as OP).
```
Without the comma it makes it sound like:  `283 of the 264 are in their own post`

FoxxMD:
Add a comma to clarify that comment total is an introductory phrase. Clarifies that the proceeding clause is specifying a subset of the total, rather than the total being a subset of the proceeding number.

Fixes #419

### DIFF
--- a/extension/data/modules/historybutton.js
+++ b/extension/data/modules/historybutton.js
@@ -152,7 +152,7 @@ function historybutton () {
                                 <label class="submission-count"></label> submissions
                                 <br>
                                 <span class="tb-history-comment-stats" style="display:none">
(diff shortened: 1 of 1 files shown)


Geo1088 (approved):

Geo1088:
Thanks for the fix!

PR #424, opened 2020-12-09
Use $.data() less

### LINKED ISSUE #422: Toolbox breaks if subreddit name is all numbers
Probably jQuery shenanigans again...

Geo1088:
`$.data()` tries to read from jQuery's data cache or from `data-*` attributes. When it does the latter, it tries to process attribute values as non-string values if it thinks that's appropriate. It's rarely right.

Fixes #422 - using `$.data('subreddit')` on elements with a `data-subreddit` value with a subreddit name that's numeric was doing the wrong thing. `$.attr('data-subreddit')` always returns a string (and skips the jQuery data cache, which is also good for clarity.)

We should try to make an effort to remove other uses of `$.data()` as well, but this is all I'm gonna do for this PR.

### DIFF
--- a/extension/data/modules/config.js
+++ b/extension/data/modules/config.js
@@ -337,7 +337,7 @@ function tbconfig () {
 
         // If it is one of the many buttons on a queue page we first have to fetch the data and see if it is there.
         $body.on('click', '#tb-config-link, .tb-config-link', function () {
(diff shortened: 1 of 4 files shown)


creesch:
LGTM, I honestly thought we already made an effort once to remove usage of it a few years ago exactly because of this sort of stuff...

PR #426, opened 2020-12-09
Fix usernotes manager buttons not working

Geo1088:
Close button not working probably caused by #380. We really need a helper for these guys, so much different markup everywhere...

Other buttons not working probably an oversight from #223, which is... very old by now. Maybe #380 had a hand in it too, not sure how else this bug would've lasted that long.

### DIFF
--- a/extension/data/modules/usernotes.js
+++ b/extension/data/modules/usernotes.js
@@ -849,7 +849,7 @@ function usernotes () {
             });
 
             // Update user status.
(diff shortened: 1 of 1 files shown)


creesch:
> Other buttons not working probably an oversight from #223, which is... very old by now. Maybe #380 had a hand in it too, not sure how else this bug would've lasted that long.

Probably because we haven't had a release in a while and this code isn't live yet? We probably should work towards releasing our current code.

PR #430, opened 2020-12-13
Favor browser console filtering tools for logging

### LINKED ISSUE #96: Rework debugging console
On discord we discussed reworking the debugging console, I figured I'd make an issue here so it is all documented. 

## Do away with the html console

I rarely use it, it is heavy and cumbersome. I'd like to rework the debug window to just have the controls for console output. 

If it turns out we want a dedicated console I propose we make use of a background page where console messages get send to so we can use the browser console there. 

## Get rid of the eval method

Frankly it is sometimes useful but it is also a security risk that slows down review. 

## Move debug console code to tbDebug.js or something like that.
It currently is a jquery plugin, which is silly. I'd like to make it a proper class in itself that can be called for outputting messages and given filter parameters.

Due to the way logging works in toolbox this is also fairly easy to do. We need to replace all `$.log` instances with the new method of which there aren't a whole lot (around 60) and map the module `self.log` to that as well for initial migration. We can then refine how `self.log` works as well later on.

## Use console.trace in combination with a collapsed group, possibly also add timestamp.
Currently it is impossible to see from where a log function was called. Putting the console message in a collapsed group that includes a console.trace() would give better insights. Example:

```javascript
console.groupCollapsed('[caller]', message);
console.trace()
console.groupEnd()
```

As the headline says, possibly add a timestamp. 

```javascript
console.log(new Date().toISOString());
```

console.log(new Date().toISOString());

## Have various color coded types 

```javascript
console.log('[Caller] %c Error:', 'background-color: #e80303; color: #fff', 'This is a serious error');
```

outputs

image

I want to basically mimic native console types but with our own coloring. Doing it like this allows us to filter just toolbox related things. 

## Better filter options 
I'd like to be able to define a variety of filters from toolbox itself. 

- More refined module selection. Currently it is a dropdown, I'd like it to be a selection list so we can show multiple modules at once. 
- Filter on message type.

Geo1088:
Since we moved toolbox's logging to the browser console, it makes more sense to just use the built-in console filtering tools present in Chrome and Firefox rather than building something ourselves.

Log levels are now associated with a `console.*` function that sets their level in the console. Because of this, the old method of using `console.group` to hide a stack trace is no longer used; logs where the stack trace matters should use `log.error`, which always emits a traceback, or `log.warn`, which emits a traceback on Chrome but not Firefox.

I've also removed timestamp display, as both Chrome and Firefox have their own options for adding timestamps to messages, and adding it otherwise just looks kinda cluttered. In its place, the string `tb:` is at the beginning of every log message, making it easy to filter for only Toolbox's log messages.

Worth noting that in Chrome, the default message filters do not include the `debug` type, but in Firefox they do. This means that Firefox console transcripts from users will contain lots of useless messages. I don't think this is a major issue, though, and it's easy enough to instruct a user to toggle off the "Debug" button in the console window if necessary.

Fixes #96 (finally!)

### DIFF
--- a/extension/data/tblog.js
+++ b/extension/data/tblog.js
@@ -3,39 +3,30 @@
 (function () {
     /** @module TBLog */
 
-    // The various log types by name, and their colors applied via CSS
+    // Settings for each log type, by name
     const logTypes = {
         debug: {
             color: '#fff',
             background: '#387fa780',
+            func: console.debug,
         },
         info: {
             color: '#fff',
             background: '#38a76280',
+            func: console.info,
         },
         warn: {
             color: '#fff',
             background: '#ce821a80',
             text: 'warning',
+            func: console.warn,
         },
         error: {
             color: '#fff',
             background: '#eb394180',
+            func: console.error,
         },
     };
-
-    // Objects recording which callers and log types are being filtered
-    const filteredTypes = {};
-    const filteredCallers = {};
-
-    /**
-     * Generates a short timestamp for the current time. Strips unneeded info
-     * from the ISO format.
-     * @private
-     */
-    function timestamp () {
-        return new Date().toISOString().replace('T', ' ').replace(/Z|\..*/, '');
-    }
 
     /**
      * Executes a log of a given type.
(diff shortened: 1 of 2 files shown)


extension/data/tblog.js
```diff
@@ -104,16 +87,11 @@
...
-    TBLog.filterType = type => filteredTypes[type] = true;
-    TBLog.unfilterType = type => delete filteredTypes[type];
-    TBLog.filterCaller = caller => filteredCallers[caller] = true;
-    TBLog.unfilterCaller = caller => delete filteredCallers[caller];
-
     // Huzzah!
     window.TBLog = TBLog;
 })();
+
+TBLog.debug('test');
```
creesch, line 94: forgot something?

creesch:
Generally looks good to me except that tracing is now missing for everything except errors. I still have to testdrive this but the reason I wanted the stacktraces included was not just for errors but to be able in general to trace exactly back where a log message came from as that isn't always clear with regular messages. 

It also isn't that big of a deal though as I don't think I have used it that much anyway so I am not going to let it be a showstopper.

Geo1088:
I haven't used the traceback groups at all, which is why I went ahead and removed it. The fact that `console.group` group headers don't have a level and can't be filtered was just too much of a pain, and I don't think tracebacks are really that important unless the code hits an error condition that needs to be better understood - in which case `log.error` is a more appropriate level anyway.

Once this gets merged I'll be working on using more appropriate levels for things of that nature where I find them.

PR #439, opened 2020-12-30
Show ignored reports

### LINKED ISSUE #207: Make the "show ignored reports" feature actually work.
https://www.reddit.com/r/toolbox/comments/ec8o01/viewing_ignored_reports_appears_to_be_broken/

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbapi.js#L656

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/4bf794a2e942e21b1706aaebee0a9f75baadfe6b/extension/data/modules/queuetools.js#L262

Needs to make use `user_reports_dismissed`.

eritbh:
Fixes #207. Marking "enhancement" because this somewhat widens the scope of this feature.

I've rewritten the "Show report reasons" queue tools option from scratch. I got rid of `TBApi.getReportReasons` and instead added report-related information to the return value of `TBCore.getApiThingInfo`. Then, I rewrote the UI to use the frontend API, making it accessible from new Reddit as well.

Screenshot

### DIFF
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -1423,7 +1423,7 @@ Action reason: ${value.data.details}
                 // Create the button and add its event listener
                 const $button = document.createElement('a');
                 $button.classList.add('tb-bracket-button');
-                $button.textContent = 'reports';
+                $button.textContent = 'show reports';
                 $button.addEventListener('click', clickEvent => {
                     // Construct the list of reports
                     const reportList = document.createElement('div');
--- a/extension/data/modules/queuetools.js
+++ b/extension/data/modules/queuetools.js
@@ -864,7 +864,7 @@ function queuetools () {
                         TB.ui.longLoadNonPersistent(false, 'Sorting sidebar...', TB.ui.FEEDBACK_NEUTRAL);
                         $sortButton.html('sort by items');
                         $sortButton.css({'padding-left': '', 'padding-right': ''});
-                    }
+                    },
                 );
 
                 function sortSubreddits () {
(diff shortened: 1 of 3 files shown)


extension/data/modules/queuetools.js
```diff
@@ -1432,6 +1397,77 @@ Action reason: ${value.data.details}
...
+                const {id, author} = redditEvent.detail.data;
+                const {reportsIgnored, userReports, modReports} = await new Promise(resolve => TBCore.getApiThingInfo(id, subreddit, false, resolve));
+                if (!reportsIgnored) {
+                    return;
+                }
+
+                // Create the button and add its event listener
+                const $button = document.createElement('a');
+                $button.classList.add('tb-bracket-button');
+                $button.textContent = 'reports';
```
SpyTec, line 1426:
```suggestion
                $button.textContent = 'show reports';
```

SpyTec (commented):
Seems to work in new Reddit, but I'm not getting any reports button in old Reddit

extension/data/modules/queuetools.js
Geo1088: I intentionally made this short since it's displayed next to the author name in comments on new Reddit - don't want it to take up too much space. Do you think it's still okay like that?

Geo1088:
Addressed the changes @SpyTec suggested, and corrected a bug that made the title always say "/u/someone's post" even on comments. Ready for review again.

dequeued0:
It seems like "Ignored reports" may be potentially confusing phrasing for this dialog because of the "ignore reports" button. I'd consider calling them "dismissed reports" because that's what they are called in the JSON and I think it's a clearer and more accurate term.

Also, would it be possible to put the button for this on the same line as "show recent actions"? It seems like a waste of vertical space to put it above "show recent actions".

eritbh:
> It seems like "Ignored reports" may be potentially confusing phrasing for this dialog because of the "ignore reports" button. I'd consider calling them "dismissed reports" because that's what they are called in the JSON and I think it's a clearer and more accurate term.

Good point. I think old Reddit calls them "Old reports" in the interface, so I'll probably update it to that for consistency.

> Also, would it be possible to put the button for this on the same line as "show recent actions"? It seems like a waste of vertical space to put it above "show recent actions".

I'll have to look into this. I don't remember exactly how those buttons interact in terms of styling; we may need to implement a better way of laying out multiple buttons in this area, which would probably be better addressed in a separate PR.

creesch:
> Good point. I think old Reddit calls them "Old reports" in the interface, so I'll probably update it to that for consistency. 

This still needs to be done right?

eritbh:
Not anymore, 1576d13

PR #447, opened 2021-01-21
Correctly interpret RR settings when forced by subreddit settings

### LINKED ISSUE #445: Removal reasons settings aren't read correctly when the "force" sub option is used
https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/modules/removalreasons.js#L647-L648

When the removal reasons settings are forced on mods, the inputs for these settings are disabled to prevent them from being modified. When we read back the settings later, we ignore their values if they're disabled, probably due to a previous UI behavior where things would get disabled if prerequisites weren't selected. This means that the settings for locking, etc. will always be read as `false` if the subreddit forces the settings.

Geo1088:
Fixes #445. I think previously some of these controls were disabled if the parent setting for them wasn't selected (e.g. "sticky reply" was disabled if you were sending the reason as a PM rather than a comment, since PMs can't be stickied). Now though it doesn't seem like they're disabled unless they're set and forced by subreddit configuration, and in that scenario we want them to apply, just not be changed.


creesch:
>  Now though it doesn't seem like they're disabled unless they're set and forced by subreddit configuration

Did you do a extra check to confirm this? [Seems like you actually added it 4 years ago](https://github.com/toolbox-team/reddit-moderator-toolbox/commit/ca0d4416f389e7ebed111383246e808960b32588) specifically to deal with comments. I haven't checked it myself yet properly, but just wanted to bring it up before we introduce regression here.

Geo1088:
Ohh, that makes sense. You're right, I'll have to change this around a bit to properly handle comments.

Geo1088:
Resolved with last push. The disabled state of the inputs is still ignored when reading the settings in code, but the options that only apply to posts will not be checked for visual consistency.

In the future we might want to have a clearer distinction here between "You can't change this because it doesn't apply to the kind of thing you're removing" and "you can't change this because your subreddit enforces these settings".

SpyTec:
> In the future we might want to have a clearer distinction here between "You can't change this because it doesn't apply to the kind of thing you're removing" and "you can't change this because your subreddit enforces these settings".

In an ideal UX world we wouldn't even display it. Why show the user content when it is completely useless and is just clutter?

SpyTec (approved):
Not tested it in a browser but LGTM

PR #458, opened 2021-02-24
Changes to the flair selector

### LINKED ISSUE #457: Flair selection has visual bugs with flair template selection
Introduced in #418. There are both bugs and edge-cases not taken care of:

- [Clearing a flair doesn't work?](https://www.reddit.com/r/toolbox/comments/lqhs25/bugs_unable_to_save_blank_css_flair_class_and/)
- No way to get value from flair templates to text and class
- It is not clear if selecting a flair template will overwrite text and class

Xeoth:
Fixes #457. Besides that, adds colors to the dropdown to show how the templates will look on New Reddit.

### DIFF
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -682,7 +682,6 @@ function modbutton () {
             $textinput.val(userFlairInfo.current.flair_text);
             $classinput.val(userFlairInfo.current.flair_css_class);
 
-            console.log(userFlairInfo.current);
             if (userFlairInfo.current.flair_template_id) {
                 $classinput
                     .attr('disabled', '')
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -230,14 +230,14 @@ function modbutton () {
                         title: 'User Flair',
                         tooltip: 'Edit User Flair.',
                         content: `
-                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-text" class="mod-popup-flair-label">Text:</label><input id="flair-text" class="flair-text tb-input" type="text"></input></p>
-                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-class" class="mod-popup-flair-label">Class:</label><input id="flair-class" class="flair-class tb-input" type="text"></input></p>
                     <p style="clear:both;" class="mod-popup-flair-input">
                         <label for="flair-template-id" class="mod-popup-flair-label">Template:</label>
                         <select style="text-overflow: ellipsis; width: 150px;" id="flair-template-id-select" class="tb-action-button">
                             <option value="">None</option>
                         </select>
-                    </p>`,
+                    </p>
+                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-text" class="mod-popup-flair-label">Text:</label><input id="flair-text" class="flair-text tb-input" type="text"></input></p>
+                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-class" class="mod-popup-flair-label">Class:</label><input id="flair-class" class="flair-class tb-input" type="text"></input></p>`,
                         footer: `
                 <span class="status error left"></span>
                 <button class="flair-save tb-action-button">Save Flair</button>`,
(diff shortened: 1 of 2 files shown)


creesch:
@Geo1088 you request review yourself, do you still want to look this over?

Geo1088:
I planned to, but don't let me hold this back if you think it's good to go.

extension/data/modules/modbutton.js
```diff
@@ -230,14 +230,14 @@ function modbutton () {
...
-                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-class" class="mod-popup-flair-label">Class:</label><input id="flair-class" class="flair-class tb-input" type="text"></input></p>
                     <p style="clear:both;" class="mod-popup-flair-input">
                         <label for="flair-template-id" class="mod-popup-flair-label">Template:</label>
                         <select style="text-overflow: ellipsis; width: 150px;" id="flair-template-id-select" class="tb-action-button">
                             <option value="">None</option>
                         </select>
-                    </p>`,
+                    </p>
+                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-text" class="mod-popup-flair-label">Text:</label><input id="flair-text" class="flair-text tb-input" type="text"></input></p>
+                    <p style="clear:both;" class="mod-popup-flair-input"><label for="flair-class" class="mod-popup-flair-label">Class:</label><input id="flair-class" class="flair-class tb-input" type="text"></input></p>`,
```
Geo1088, line 240: Style nit: I know you just moved these lines down, but could you clean them up so the elements aren't all strung together on one line?

Geo1088 (changes requested):
This PR has undesirable behavior when clicking into the flair tab, then out, then back in. The flair templates end up duplicated in the dropdown, and the values of the text boxes are also overwritten with the current value each time this happens. This is probably a result of the way the mod button's UI code has been structured for a long time, rather than being a result of your specific changes, but I think we should do something about that before we merge this.

image

Is it possible to move the setup code so it happens once, when the popup is first created, rather than only when the tab is entered? If not, we should at least store some indication that the tab has already been loaded and shouldn't be set up again.

extension/data/styles/toolbox.css
```diff
@@ -530,10 +530,16 @@ body.mod-toolbox-rd.mod-toolbox-extra .tb-input:focus {
...
-.mod-toolbox-rd .mod-popup .flair-text {
+.mod-toolbox-rd .mod-popup .flair-text,
+.mod-toolbox-rd .mod-popup .flair-class {
     width: initial !important;
 }
 
+.mod-toolbox-rd .mod-popup .mod-popup-flair-input label[for="flair-template-id"] {
+    position: relative;
+    bottom: 4px;
+}
```
Geo1088, line 541: One more thing I forgot to mention: You should be able to use `vertical-align: baseline !important` to do this centering rather than relying on messy relative positioning. `.tb-action-button` gets `vertical-align: bottom !important` which you can override back to the default value.

Xeoth:
Thanks for the review, it should all be fixed now.

extension/data/styles/toolbox.css
```diff
@@ -530,10 +530,15 @@ body.mod-toolbox-rd.mod-toolbox-extra .tb-input:focus {
...
 }
 
-.mod-toolbox-rd .mod-popup .flair-text {
+.mod-toolbox-rd .mod-popup .flair-text,
+.mod-toolbox-rd .mod-popup .flair-class {
     width: initial !important;
 }
 
+.mod-toolbox-rd .mod-popup .mod-popup-flair-input label[for="flair-template-id"] {
+    vertical-align: super !important;
```
Geo1088, line 539: Is this `super` instead of `baseline` intentionally? I haven't tried running this again, if it looks good to you then we can go with it.

Geo1088: Weird issues with `baseline` discussed [in Discord](https://discord.com/channels/535490452066009090/535490452066009093/821466215381467146), using `super` is good enough for now and I'll plan to look over this more when I'm in the neighborhood with #459.

Geo1088 (approved):

PR #462, opened 2021-03-02
Add type check when purifying parsed JSON values

### LINKED ISSUE #461: Usernames in scientific notation are interpreted incorrectly somehow
https://new.reddit.com/r/toolbox/comments/ltwj4e/bug_mod_actions_dialog_does_not_work_correctly/

Been investigating this, issue seems to be caused by the `TBStorage.purifyObject` function called on the response of API requests. Which is... not great. Will post further investigation and (hopefully) a PR to fix this shortly.

Geo1088:
Fixes #461. Uses the type-check method discussed in Discord. Expands a bit on the surrounding comments to make it clearer what's going on here.

### DIFF
--- a/extension/data/tbstorage.js
+++ b/extension/data/tbstorage.js
@@ -296,10 +296,19 @@ function storagewrapper () {
                         purifyObject(input[key]);
                         break;
                     case 'string':
-                        // Let's see if we are dealing with json.
-                        // We want to handle json properly otherwise the purify process will mess up things.
+                        // If the string we're handling is a JSON string, purifying it before it's parsed will mangle
+                        // the JSON and make it unusable. We try to parse every value, and if parsing returns an object
+                        // or an array, we run purifyObject on the result and re-stringify the value, rather than
+                        // trying to purify the string itself. This ensures that when the string is parsed somewhere
+                        // else, it's already purified.
+                        // TODO: Identify if this behavior is actually used anywhere
                         try {
                             const jsonObject = JSON.parse(input[key]);
+                            // We only want to purify the parsed value if it's an object or array, otherwise we throw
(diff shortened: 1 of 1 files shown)


extension/data/tbstorage.js
```diff
@@ -296,10 +296,19 @@ function storagewrapper () {
...
+                        // or an array, we run purifyObject on the result and re-stringify the value, rather than
+                        // trying to purify the string itself. This ensures that when the string is parsed somewhere
+                        // else, it's already purified.
+                        // TODO: Identify if this behavior is actually used anywhere
                         try {
                             const jsonObject = JSON.parse(input[key]);
+                            // We only want to purify the parsed value if it's an object or array, otherwise we throw
+                            // back and purify the raw string instead (see #461)
+                            if (typeof jsonObject !== 'object' || jsonObject == null) {
+                                throw new Error('not using the parsed result of this string');
```
creesch, line 310: Why are you throwing an error here? Better would be to just assume it was a string to begin with and purify the string value and return that in my opinion.

creesch (changes requested):

extension/data/tbstorage.js
Geo1088: That's exactly what this does - the error is caught and the code in the `catch` block is executed. We use try/catch here because `JSON.parse` throws if the string isn't a JSON string; I'm just also throwing another error here if the parsed value isn't an object/array. This is explained in the comments right above. I did it this way to avoid repeating code, but I can change it if you still don't think it's clear enough.

creesch: Uh nah it is all good, I am being daft here by overlooking that.

PR #467, opened 2021-03-12
Only do ban related logic for banning

creesch:
Because it fails otherwise, this simply makes sure we only try to do things related to banning when it is needed,


Geo1088 (approved):
Huh, not sure why I only excluded the ban params from one specific call in #458. This looks good to me.

Issue #448, opened 2021-01-27
Expose `mod_note` on old reddit as well as making it more visible on redesign.


creesch:
Redesign removal reasons have the option to leave a mod note with the removal reason

![image](https://user-images.githubusercontent.com/2502190/106037857-80331a80-60d7-11eb-9303-8b840d572451.png)

This note is then exposed through the `thing` json data as the `mod_note` property. 

I did some little research today and while it isn't documented there is an endpoint for it under `/api/v1/modactions/removal_reasons` which takes the following payload

```json
{
        "item_ids": ["t3_etc"],
        "mod_note": "this is text",
        "reason_id": null,
}
```

From my testing it only seems to work on removed things but can be applied to both comments and posts.

I'd like to expose this piece of information with toolbox in a variety of ways: 

- Show it under comments and posts that have been removed on old reddit.
- Show it more obviously on redesign (currently it is hidden under a hover for the text "removal reason")
- Add it as an option to our own removal reasons

Before we can do that our current request model needs a slight adjustment as the endpoint expects a JSON body and not formdata.

creesch:
Came across this today and some things came to mind: 

- I can't remember if it was just for posts or also for comments, need to verify that. 
- Should adding a note only be possible through removal reasons or also have a separate interface?
- I am thinking that we should look it up for each removed things we encounter and then cache it for a certain period.

PR #468, opened 2021-03-14
Improve request interface

eritbh:
Adds support for passing plain string bodies for POST requests, rather than always interpreting `options.body` as form data. This clears the way for features using new APIs that use JSON bodies rather than form data, such as #448. Also simplifies the code used to send messages from the background page back to the content script when a request has finished, and corrects some documentation comments.


creesch:
Is there any particular reason this isn't merged? Or is it just awaiting review?

eritbh:
Waiting for review, and like most of my open PRs I wanted to land them after we're done with 5.5.x releases - but we seem to keep needing more bug fix releases :P

PR #477, opened 2021-03-26
Implement automated updating of DOMPurify fixes #475

### LINKED ISSUE #475: Use github actions to update dependencies (specifically domPurify)
Mozilla doesn't like it when we don't include the latest release of domPurify. Which effectively blocks any plans for automated releases on our part as we don't use NPM dependencies inside toolbox so we are not alerted through a nice automated PR about updates. 

The solution I have in mind is to this ourselves through github actions. 

1. Schedule automated running of an update workflow.
2. Have it pull our master
3. Have it pull the [latest](https://github.com/cure53/DOMPurify/releases/latest) domPurify release (specifically `/dist/purify.js`) 
4. As we don't have easily accessible metadata available (I don't think you can store persistent data in actions) about the current version simply do a string equals or hash check on the contents of our version and the latest one. 
5. If a difference is found overwrite our version with the pulled version. 
6. [Create an automated PR](https://github.com/marketplace/actions/create-pull-request) 

Step 4 is heavily based on assumptions, if we can actually do an actual version check that would have my preference.

creesch:
Fixes #475 through the use of [this awesome github action someone wrote](https://github.com/creesch/github-latest-release-zip) and the [git pull request action](https://github.com/peter-evans/create-pull-request).  

It works as discussed on discord without actual logic for version checking by simply pulling the latest release, copying over the file and then seeing if there is a delta for which a pull request can be done. 

I did most of the testing on a dead repo, an example of a created PR can be [found here](https://github.com/creesch/jsapi-example-consumer/pull/8)

### DIFF
--- a/.github/workflows/dependency_update_check.yml
+++ b/.github/workflows/dependency_update_check.yml
@@ -1,0 +1,44 @@
+name: Check for DOMPurify updates
+
+on:
+  schedule:
+    # run daily at midnight
+    - cron: '0 0 * * *'
+jobs:
+  checkdepedencies:
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout toolbox
+      - name: checkout
+        uses: actions/checkout@v2
+      # Download latest DOMPurify
+      - name: download
+        id: download
+        uses: creesch/github-latest-release-zip@main
+        with:
+          owner: cure53
+          repo: DOMPurify
+          downloadPath: build
+      # Unzip and copy over the release purify.js if there is an update this will create a delta.
+      - name: unzip
+        run: |
+          BASEDIR=$(pwd)
+          cd build
+          FILENAME="${{ steps.download.outputs.filename }}"
+          unzip $FILENAME
+          DIRECTORY=$(ls -d */)
+          cp "${DIRECTORY}dist/purify.js" "${BASEDIR}/extension/data/libs/purify.js"
+      # Run pullrequest action, if the previous stepped caused no difference no pullrequest will be made.
+      - name: pullrequest
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: Update DOMPurify to ${{ steps.download.outputs.tag_name }}
+          title: Update DOMPurify to ${{ steps.download.outputs.name }}
+          body: |
+            [${{ steps.download.outputs.name }}](${{ steps.download.outputs.html_url }})
+
+            ## Release notes:
+
+            ${{ steps.download.outputs.body }}
+          branch: update-DOMPurify-dependency
+          delete-branch: true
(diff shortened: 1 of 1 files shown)


eritbh (approved):
LGTM! Elegant solution - very nice that `create-pull-request` won't error out if there's no local changes.

Not sure I entirely follow your bash code (what happens to the other files in the zip? do they not make it to the PR?) but if this makes wrong output then we can just reject the PR it generates, so I think this is good to merge - we can tweak it later as needed.

creesch:
> Not sure I entirely follow your bash code 

Afaik we aren't starting out in root with a environment so I set the current dir here.
```
BASEDIR=$(pwd)
```
The zip was downloaded to the `build` dir. 
```
 cd build
```

Using `FILENAME` as a variable is probably not really needed but basically we unzip the zip here.
```
FILENAME="${{ steps.download.outputs.filename }}"
unzip $FILENAME
```

Unzipping is done to a directory for which we want to know the name otherwise we can't copy stuff. As we know that there is no other directory in the current directory we can just do `ls` with a little trickery to find the only directory available. 
```
DIRECTORY=$(ls -d */)
```

Now we know the unzipped directory we can copy purify.js over to the extension dir. 
```
 cp "${DIRECTORY}dist/purify.js" "${BASEDIR}/extension/data/libs/purify.js"
```

Now to answer your actual question

> (what happens to the other files in the zip? do they not make it to the PR?) 

`build` is a gitignored directory zo the only file that ever is visible as far as git is concerned is `purify.js`.

.github/workflows/dependency_update_check.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: checkout
+        uses: actions/checkout@v2
+      # Download latest DOMPurify
+      - name: download
+        id: download
+        uses: creesch/github-latest-release-zip@main
+        with:
+          owner: cure53
+          repo: DOMPurify
+          downloadPath: build
```
eritbh, line 21: We should probably add some comments to mention the significance of using `build` (a gitignore'd directory) as the download destination.

.github/workflows/dependency_update_check.yml
```diff
@@ -0,0 +1,44 @@
...
+      # Unzip and copy over the release purify.js if there is an update this will create a delta.
+      - name: unzip
+        run: |
+          BASEDIR=$(pwd)
+          cd build
+          FILENAME="${{ steps.download.outputs.filename }}"
+          unzip $FILENAME
+          DIRECTORY=$(ls -d */)
```
eritbh, line 29: Why not use `zip -d` to explicitly set the output directory? I see why this works, but I think using `-d` might be a bit easier to understand. (I try to avoid command substitution when possible when I write shell scripts, but that might just be a me thing.)

eritbh (commented):
Gotcha. I was slightly confused about how the `build` directory was present after cloning since it's gitignored, but I didn't notice that that's where the file is being downloaded to. Normally for stuff like this I'm used to seeing actions doing `mkdir ../temp` or something similar and putting all their temporary files there, but this works too.

eritbh (approved):

PR #478, opened 2021-03-27
Don't use /api/flairselector for user flair info

eritbh:
Rather than hitting `/api/flairselector` to get information about the user's current flair, we just get this information from the post info we're triggering the mod button on.

### DIFF
--- a/extension/data/modules/modbutton.js
+++ b/extension/data/modules/modbutton.js
@@ -274,7 +274,7 @@ function modbutton () {
             $popup.find('.tb-window-tabs .user_flair')
                 .attr('data-flair-text', info.author_flair_text || '')
                 .attr('data-flair-css-class', info.author_flair_css_class || '')
-                .attr('data-template-id', info.author_flair_template_id || '');
+                .attr('data-flair-template-id', info.author_flair_template_id || '');
 
             const $actionSelect = $popup.find('.mod-action');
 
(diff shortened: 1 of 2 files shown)


extension/data/modules/modbutton.js
```diff
@@ -673,30 +681,29 @@ function modbutton () {
...
-            if (!userFlairInfo.current) {
-                return;
-            }
-
-            TBStorage.purifyObject(userFlairInfo);
-            $textinput.val(userFlairInfo.current.flair_text);
-            $classinput.val(userFlairInfo.current.flair_css_class);
+            const currentFlairText = $this.attr('data-flair-text');
+            const currentFlairClass = $this.attr('data-flair-css-class');
+            const currentFlairTemplateID = $this.attr('data-flair-template-id');
```
Xeoth, line 686: This should be `data-template-id`. Or we could change the context to `data-flair-template-id` for consistency. Either way, it doesn't pick up the template ID right now.

eritbh:
Also requesting review from @creesch - this modifies `getApiThingInfo` in a way that makes it incompatible with `getThingInfo`. The properties I've added aren't probably useful anywhere else, and they seemed like a pain to pull from DOM on old Reddit, but want to make sure you think it's okay too.

creesch:
It looks sensible to me, my only question is if this also works properly in modmail context?

eritbh:
Closing - this method doesn't work in modmail. Don't know of any alternative methods to get a user's current flair, so I'll post to redditdev about this endpoint returning the wrong thing and we'll see what we can do from there.

Issue #174, opened 2019-10-15
queue sorting setting might not work


creesch:
https://www.reddit.com/r/toolbox/comments/dc5yiz/is_there_a_way_to_sort_the_modqueue_so_that_the/f3ucqxg/

Issue so we don't forget.

k3n:
I'm going to keep an eye on this (since that was me that reported it).

PR #482, opened 2021-03-29
Fix age sorting in queues

creesch:
Using `live-timestamp` as a selector should be fine under normal circumstances except that reddit doesn't always apply that class to queue items for reasons unknown to me (did test it with no extensions enabled just to be sure). I am fairly sure it started happening recently but as I don't want to rely on reddit fixing it I decided to adjust the selectors in such a way that this shouldn't (hopefully) cause any issues moving forward. 

Could use a look over to see if I didn't mess anything up in my thinking here.


creesch:
Also this might have been the same issue reported in #174 as reported by @k3n suggesting that this is a recurring issue on reddits side (I am guessing caused by backend weirdness or simply regression).

eritbh (approved):

PR #484, opened 2021-03-30
ES6 modules

eritbh:
Start of #198.


eritbh:
I think as of my last push, I've converted all the files I reasonably can before we start redefining the module structure. Before I do that, we need to decide our strategy for further work on this. As I see it, we either get all the work done at once in one PR, or we merge what we have now and start the next phase in a separate PR. If we merge now, ESLint is going to be broken for a lot of things because it doesn't have a good way of handling a mix of source types (script vs. module). If we keep working in this branch and merge at the end, we can fix ESLint at the same time as all the other changes, but we'll essentially have to freeze all other changes to the codebase until we're done or risk creating large, ugly conflicts.

@creesch let me know if you think the direction I'm going with this code looks good, and what you think our next move should be once we're happy with the state of these changes.

eritbh:
At this point this branch has module-ify'd basically everything except for `TBCore`, which I'll have to rework a bit more deeply because of its reliance on `userDetails` and other values that aren't obtained synchronously. Basically everything that relies on those values will have to become asynchronous so we can guarantee the values are present before trying to do anything with them. This will have a big impact on other files, because TBCore is so ubiquitous. I've considered reworking TBCore into a class for this reason, but then the issue becomes how to share one instance of that class between all modules - it doesn't really solve the issue of asynchronously obtained data that's necessary to initialize the module.

Note to self: Implementing this will be similar to how I've already done `TBStorage.getSettingAsync` and `TBStorage.setSettingAsync`, with a promise that fetches data from the background page and asynchronous functions that wait for that promise to fulfill before doing anything, but slightly different since actually fetching the user details is a function of the init process rather than TBCore itself. TBCore constant properties that rely on `userDetails` will probably become functions that return promises, but other constants can be exported as-is.

creesch:
Damn you have been busy :D Sorry I haven't been able to put in more time to support you with this. 

 In regards to TBCore, I already did split a lot of things out of it to TBHelpers and TBApi but it might be worth it to see if we can separate the things that require the async data to be there from the things that do not require them. That might make it easier to then find a suitable solution for the async issue. 


Also, so far you have been working with the import model supported by extensions. The plan is still to eventually switch to a build tool right? I do understand not doing that at the same time or even in this branch, just making sure we are on the same page.

eritbh:
>  it might be worth it to see if we can separate the things that require the async data to be there from the things that do not require them.

Yep, I have a couple more local commits where I've basically switched to that strategy. I've also been going through and trying to find properties of `TBCore` that were essentially just working as proxies to stuff stored in `TBStorage` settings/cache, replacing those with direct storage calls to simplify things and avoid storing the same state in multiple locations.

> Also, so far you have been working with the import model supported by extensions. The plan is still to eventually switch to a build tool right? I do understand not doing that at the same time or even in this branch, just making sure we are on the same page.

The syntax I'm adopting now is supported in Chromium-based browsers, but won't be fully supported in Firefox until FF 89 is available as an extended support release. The goal of using a bundler would be to convert this syntax to something that Firefox can work with until then, in order to push this earlier. A bundler would also allow us to stop storing dependencies in a `libs` folder and install/update them with npm instead, but I'm not interested in doing that as part of this PR - that's super easy to deal with after the fact if we want it.

eritbh:
(I should probably just squash this branch's history, this is getting a bit ridiculous...)

Almost all of `TBCore` is converted now. All functions/constants are exported with the ES6 syntax, and only a handful of values are still added as properties to the `window.TBCore` object. Once that's done I think this will be ready for... whatever sort of review can be done on a change set this big. :V

I haven't done the `TBModule` refactoring that I mentioned in #198, but I think that can be done easily with a separate PR once this is done. Reworking our asynchronous init process to play nicely with ES modules is all I want to do on this branch, and I'm nearly done with that.

package.json
```diff
@@ -4,8 +4,7 @@
     "description": "A set of tools to be used by moderators on reddit in order to make their jobs easier.",
     "private": true,
     "scripts": {
-        "build": "node build.js",
-        "build-watch": "nodemon --exec npm run build",
+        "build": "rollup -c rollup.config.js",
```
creesch, line 7: No `build-watch` anymore? ;)

eritbh: Rollup has a built-in watcher, just `npm run build --watch` :)

eritbh:
Superseded by #518

PR #487, opened 2021-04-07
fix issue with overflowing removal reasons

creesch:
Fixes a rare issue where removal reasons containing a long first line with no spaces would cause the removal reason config tab to overflow.

### DIFF
--- a/extension/data/styles/config.css
+++ b/extension/data/styles/config.css
@@ -33,7 +33,7 @@
 /* removal reasons */
 .mod-toolbox-rd .removal-reasons-content .removal-reason-label {
     white-space: break-spaces;
-    word-break: break-all;
+    overflow-wrap: anywhere;
 }
 
 /* ban-macro */
--- a/extension/data/styles/config.css
+++ b/extension/data/styles/config.css
@@ -30,6 +30,12 @@
     z-index: 9999999999;
 }
 
+/* removal reasons */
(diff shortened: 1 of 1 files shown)


extension/data/styles/config.css
```diff
@@ -30,6 +30,12 @@
     z-index: 9999999999;
 }
 
+/* removal reasons */
+.mod-toolbox-rd .removal-reasons-content .removal-reason-label {
+    white-space: break-spaces;
+    word-break: break-all;
```
eritbh, line 36: Not tested locally, but are you sure `word-break: break-all` is what you want here? [MDN seems to suggest](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#values) this will result in words getting broken even if they don't need to be:

> ...`word-break: break-all` will create a break at the exact place where text would otherwise overflow its container **(even if putting an entire word on its own line would negate the need for a break).**

You might want [`overflow-wrap: break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) instead, which will only separate words if necessary.

creesch: Fairly sure, I have tried a variety of css properties to make it wrap properly. I'll double check later today.

creesch: Somehow using `overflow-wrap: break-word` didn't achieve the desired result, but using `overflow-wrap: anywhere` does so I switched to using that.

eritbh (approved):

PR #505, opened 2021-06-30
Accept user profile permalinks

### LINKED ISSUE #504: Context popup doesn't work on comments in user profiles
Request being made to an incorrect URL - the double slashes in there makes it seem like we're not picking up the parent submission ID correctly. Seem in old Reddit.

image

We might also want to add the endpoint where the request failed to error logs for these situations, would make debugging much easier when the network panel in-page doesn't reflect any of Toolbox's requests anymore.

eritbh:
Untested, but theoretically fixes #504. Adjusts permalink parsing to accept user profile permalinks, resulting in post IDs being reported correctly through the old Reddit frontend API shim.


creesch:
I could have sworn we already did fix this at one point. Ah well, seems like we didn't. PR looks good to me so I'll merge it.

PR #514, opened 2021-07-13
Make all requests except mass actions bypass ratelimits

eritbh:
The conclusion 5.6 taught us is basically that keeping track of Reddit's rate limits strictly just isn't worth it most of the time, because it still handles requests over the limit gracefully, and too many users are having issues with responsiveness when strictly respecting limits imposed on the auth token we pull from the page. This change makes `bypassRatelimit` default to `true` when sending requests, and only explicitly sets it to `false` for the mass requests performed in comment nuke and the mod button.

In the future, if we want to strictly follow limits across the board, we'll first need to minimize our own usage of API requests as much as possible, and we'd also probably want to look into transitioning to our own OAuth-based authorization rather than grabbing credentials from the page. These changes may not (read: probably won't) be worth it, so most of the new ratelimit code isn't going to be used 99% of the time, but I guess that's how it goes.

### DIFF
--- a/extension/data/tbapi.js
+++ b/extension/data/tbapi.js
@@ -425,7 +425,7 @@
             }
         }
 
-        return TBApi.apiOauthPOST('/api/friend', {
+        return TBApi.post('/api/friend', {
             api_type: 'json',
             uh: TBCore.modhash,
             type: action,
@@ -437,7 +437,7 @@
             ban_context: banContext,
         }, {
             bypassRatelimit,
-        }).then(response => response.json());
+        });
     };
 
     /**
@@ -495,7 +495,7 @@
      * @param {boolean} [bypassRatelimit=true] Passed to TBApi.sendRequest
      * @returns {Promise}
      */
-    TBApi.removeThing = (id, spam = false, bypassRatelimit = true) => TBApi.apiOauthPOST('/api/remove', {
+    TBApi.removeThing = (id, spam = false, bypassRatelimit = true) => TBApi.post('/api/remove', {
         uh: TBCore.modhash,
         id,
         spam,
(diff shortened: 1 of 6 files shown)


extension/data/tbapi.js
```diff
@@ -477,12 +492,15 @@
      * @function
      * @param {string} id Fullname of the post or comment
      * @param {boolean?} spam If true, removes as spam
+     * @param {boolean} [bypassRatelimit=true] Passed to TBApi.sendRequest
      * @returns {Promise}
      */
-    TBApi.removeThing = (id, spam = false) => TBApi.apiOauthPOST('/api/remove', {
+    TBApi.removeThing = (id, spam = false, bypassRatelimit = true) => TBApi.apiOauthPOST('/api/remove', {
```
creesch, line 498: Can we change this to just use `TBApi.post`? It seems more people than we'd like have trouble with oauth requests.

extension/data/tbapi.js
```diff
@@ -510,11 +528,14 @@
     /**
      * Locks a post or comment.
      * @param {string} id The fullname of the submission or comment
+     * @param {boolean} [bypassRatelimit=true] Passed to TBApi.sendRequest
      * @returns {Promise} Resolves to response data or rejects with a jqXHR
      */
-    TBApi.lock = id => TBApi.apiOauthPOST('/api/lock', {
+    TBApi.lock = (id, bypassRatelimit = true) => TBApi.apiOauthPOST('/api/lock', {
```
creesch, line 534: Can we change this to just use `TBApi.post` as well?

creesch (changes requested):
Generally looks good to me but I'd like to swap out `TBApi.apiOauthPOST` for `TBApi.post` where possible as well as people also seem to have more issues with those requests.

PR #518, opened 2021-07-16
ES6 Modules for content script

eritbh:
Supersedes #484. I've squashed all the commits in the [`master...es6-modules`](https://github.com/toolbox-team/reddit-moderator-toolbox/compare/master...es6-modules) range (475406c...6a7cc83) into a new branch, which will hopefully keep things a bit more sane for code review. If you want to check individual commits, check the original PR or the compare range mentioned above. I'll add any requested changes to this branch.

This change touches pretty much every aspect of the codebase (except the background page), so even though I'm sure I've forgotten things, here's a breakdown of the most significant changes, along with the most significant things that didn't change but probably will in subsequent PRs:

## Development workflow changes

- The extension is now built via Rollup with `npm run build` and `npm run build:watch`.
  - Though we have the capability now thanks to Rollup, we actually **aren't yet pulling any dependencies from npm**. When we eventually do move to using npm to manage our dependencies, we'll also add a `prebuild` script or something like that to prevent us from building releases with outdated dependencies.
- A new ESLint plugin, `eslint-plugin-import`, was added to allow static analysis of imports and point out imported values that don't exist and other fun stuff like that.

## Feature changes

- Dev mode has been removed, because it didn't do much and it was easier to remove it than port its functionality because of the way it was implemented in TBCore.
  - Things that were previously hidden behind dev mode are now hidden behind debug mode.
    - [ ] TODO: Add settings in the devtools module for these behaviors (showing all achievements, displaying hidden settings, etc) rather than enabling them all with the debug mode switch
  - Notes from /r/tb_dev are no longer read at all.

## Architecture changes

- The entry point of our content script is a new file, `init.js`, which combines all of the init processing that was previously scattered throughout TBCore, TBModule, and other files. It performs initial checks, verifies that everything required to run is present, then pulls in and runs feature modules.
  - Notably, modules now load almost immediately, without waiting for TBStorage to be loaded. This has implications for certain modules which dynamically set their settings; see the API changes section for details.
- Previously, helpers like `TBHelpers` and `TBStorage` would do asynchronous assignment to their own properties in order to make slow-loading data available to modules after the page loads. Module exports don't support being reassigned asynchronously in that way, so for the most part, this has been replaced with function calls that return promises.

The background page has not been touched in this PR. The background page and content script share no code, so I decided to dedicate this PR only to the content script and handle the background page in a subsequent PR.

## Internal API changes

- The TBModule `Module` interface has been completely redone, and existing modules adapted.
  - The `Module` constructor takes most of the module's options (name, default enable state, settings, etc), as well as its initializer function, as arguments.
    - Some modules set the default value or other properties of their settings dynamically, based on the values of other settings (either theirs or those of other modules). Because modules are now loaded before TBStorage can return values synchronously, some setting properties can now be set by a function which will be executed as needed. For example, see [the Usernotes `subredditColorSalt` setting being hidden unless the `subredditColor` setting is enabled](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/518/files#diff-be2414fc9358fcd61e1a9247b2afb597136218e93b30c9bc265a6bb7bafe48fdR51).
  - `Module#setting(id[, value])` has been replaced with `Module#get(id)` and `Module#set(id, value)` for retrieving and storing settings. Both of the new functions are asynchronous; module code was adapted for this change.
  - In #198 I mentioned that I'd like to move away from having module-specific helper functions assigned as properties of the modules themselves, swapping to explicitly exporting helpers that should be public and simply using function declarations for private helpers. For the most part I haven't done that; in this branch I'm mostly concerned with getting things working the way they did before, so I'm trying to avoid making more changes to module code than I already have.
- Most page data which was made available through properties of TBCore (modhash, user details, etc) are now available through asynchronous functions instead.
  - [x] TODO:There are still some holdovers that are assigned asynchronously to a global `window._TBCore` object. These were left the way they were because they seemed like a pain to decouple; they should be reviewed again and ported if possible. This impacts certain high-use functions; for example, `getThingInfo` cannot be used without explicitly checking that TBCore has loaded the current user's moderated subreddits, because calling the function while that info is missing will result in an error. The upshot is that we're gonna have to make a lot more currently synchronous functions asynchronous to convert the rest of this.


eritbh:
I think I'm happy with the state this is in now. I haven't completely eliminated `window._TBCore` because `TBCore.getThingInfo` relies on synchronous access to certain things, and making it asynchronous would be so impactful that I'd be much more comfortable addressing that in its own PR that can be properly reviewed.

[This new deepscan issue](https://deepscan.io/dashboard/#view=project&pid=7592&iid=90652584) can probably be ignored for now - we should reconsider whether we're really doing beta releases or not, but that can be its own discussion.

extension/data/tbmodule.js
```diff
@@ -1,213 +1,266 @@
...
-                        }
+import TBLog from './tblog.js';
+import * as TBStorage from './tbstorage.js';
+import * as TBui from './tbui.js';
+import * as TBHelpers from './tbhelpers.js';
+import TBListener from './tblistener.js';
+import * as TBCore from './tbcore.js';
+import * as TBConstants from './tbconstants.js';
+
+// This is here to fix a circular dependency issue if it's put in the syntax highlighter module
```
SpyTec: Circular dependency of what?

eritbh: `tbmodule.js` can't import `syntax.js` because `syntax.js` already imports `tbmodule.js`. In order for this constant to be available in both files, it had to be moved here instead.

SpyTec: I don't get it though, nothing is using that const anywhere except within `syntax.js`

eritbh: It's also used in `tbmodule.js` here. The theme select is displayed both on wiki edit pages and also in the settings window. https://github.com/toolbox-team/reddit-moderator-toolbox/pull/518/files#diff-5facd4242ceb776c6c705aa1e27ff6f1264879be3e5ac2043dc30882a18a240aR687

eritbh: Moved to `tbconstants.js` in 24b7a3b

PR #520, opened 2021-07-17
Enhance typing indicator & new message visibility

creesch:
Basically what the title says. That they added typing indicators and a message bubble is great but they did so in a way that seemed to have preferred looks over functionality as they sometimes are easy to miss. So this adds a few things for better visibility.

### DIFF
--- a/extension/data/styles/newmodmailpro.css
+++ b/extension/data/styles/newmodmailpro.css
@@ -172,7 +172,7 @@ html.tb-nightmode,
 }
 
 .Thread__newMessageIndicatorBubble {
-    border: solid #47d160 2px;
+    border: solid #46d160 2px;
 }
 
 /* misc fixes.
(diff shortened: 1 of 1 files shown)


eritbh:
I feel like this should probably be a setting. Since the base feature is native now, could we just repurpose/replace the existing `checkForNewMessages` setting?

extension/data/styles/newmodmailpro.css
```diff
@@ -161,6 +161,20 @@ html.tb-nightmode,
...
+*/
+
+.ThreadViewer__typingIndicator .ThreadTypingIndicator
+{
+    outline: solid #46d160 2px;
+    padding: 2px;
+}
+
+.Thread__newMessageIndicatorBubble {
+    border: solid #47d160 2px;
```
eritbh, line 175: nit: color consistency

```suggestion
    border: solid #46d160 2px;
```

eritbh (approved):

PR #522, opened 2021-07-31
fix: don't confirm removal on shift

SpyTec:
When removing by holding shift to skip the removal reasons popup, the default Reddit behaviour kicks in. Since the user is holding shift, we don't want the "are you sure" confirmation on old Reddit to still be there, instead we should be clicking yes for them.

This moves the check for shift after the user has clicked remove and before fetching any data for popup


SpyTec:
Confirmed to be working as expected. No longer asks for confirmation when shift-clicking remove

PR #525, opened 2021-08-01
Remove now-native modmail features

### LINKED ISSUE #521: Remove redundant features from new modmail pro
Thought I was gonna do this two days ago but didn't so here's an issue for it.

Reddit has implemented typing indicators in new modmail, making the "check for new messages before sending" feature obsolete. There's also a built-in advanced search interface which means we don't need a help button for search queries anymore.

eritbh:
Fixes #521

### DIFF
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -32,12 +32,6 @@ function newmodmailpro () {
         title: 'Open modmail in nightmode',
     });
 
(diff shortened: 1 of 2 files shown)


creesch:
Hi, I am creesch and I am just testing.

creesch:
More serious review note, we probably want to remove the setting values as well. 

https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbcore.js#L355

Issue #534, opened 2021-09-20
usernotes: ability to search the note type as well


creesch:
From modmail:

> When searching through usernotes it would be useful to also include the note types in the search, especially after creating custom note types that include the main category that we want to search for and then note itself is just specific details.

PR #546, opened 2021-12-11
Filter by usernote tag

dequeues:
Implement #534 

There is now a dropdown on the usernote search page of all the user note titles, and "All" which you can filter the usertags with.
First PR here, please be kind. 👋

### DIFF
--- a/extension/data/modules/usernotes.js
+++ b/extension/data/modules/usernotes.js
@@ -1045,6 +1045,10 @@ function usernotes () {
                     <div class="tb-un-info">
                         <span class="tb-info">There are ${userCount} users with ${noteCount} notes.</span>
                         <br> <input id="tb-unote-user-search" type="text" class="tb-input" placeholder="search for user"> <input id="tb-unote-contents-search" type="text" class="tb-input" placeholder="search for note contents">
+                        <select name="tb-un-filter" id="tb-un-filter" class="selector tb-action-button">
+                            <option value="all" default>All</option>
+                            ${colors.map(op => `<option value=${op.key}>${op.text}</option>`).join('')}
+                        </select>
                         <br><br>
                         <button id="tb-un-prune-sb" class="tb-general-button">Prune deleted/suspended profiles</button>
                     </div></br></br>
@@ -1082,6 +1086,7 @@ function usernotes () {
             // Variables to store the filter text
             let userText = '';
             let contentText = '';
+            let filterKey = 'all';
 
             // Creates a new pager with the correct filtered items and replace
             // the current one with the new one, debounced because typing delay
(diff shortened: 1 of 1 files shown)


extension/data/modules/usernotes.js
```diff
@@ -1104,7 +1109,7 @@ function usernotes () {
                     }
 
                     return true;
-                });
+                }).filter(user => user.notes.filter(notes => filterKey === 'all' || notes.type === filterKey).length > 0);
```
eritbh, line 1112: What you have here works perfectly fine, we can probably get better performance out of it by putting this in the existing `notes` filter that starts on line 1099. Something like this:

```js
const filteredData = allUsers.map(user => ({
    name: user.name,
    notes: user.notes.filter(note => {
        // Filter out notes not matching `contentText`
        if (!note.note.toLowerCase().includes(contentText.toLowerCase())) {
            return false;
        }

        // Filter out note types that are not `filterKey`
        if (filterKey !== 'all' && filterKey !== note.type) {
            return false;
        }

        return true;
    }),
})).filter(user => // ...
```

Doing all the filtering in the same `.filter` call means we don't need to iterate over all the users and notes multiple times, which for usernotes can be a big concern (some subreddits have a *lot* of usernotes). This new structure will be a little more flexible and allows us to filter on multiple different properties within a single loop.

This is very minor, but I also changed the iteration variable from `notes` to `note` - inside the loop, it refers to a single note object, not an array of them, so the singular name makes the code just a bit easier to read.

extension/data/modules/usernotes.js
```diff
@@ -1045,6 +1045,10 @@ function usernotes () {
                     <div class="tb-un-info">
                         <span class="tb-info">There are ${userCount} users with ${noteCount} notes.</span>
                         <br> <input id="tb-unote-user-search" type="text" class="tb-input" placeholder="search for user"> <input id="tb-unote-contents-search" type="text" class="tb-input" placeholder="search for note contents">
+                        <select name="tb-un-filter" id="tb-un-filter" class="selector tb-action-button">
+                            <option value="all" default>All</option>
+                            ${colors.map(op => `<option value=${op.key}>${op.text}</option>`).join('')}
```
eritbh, line 1050: We unfortunately aren't super consistent about this in the rest of the code right now, but any time user input is put into HTML, it should be run through `TBHelpers.htmlEncode()` first.

```suggestion
                            ${colors.map(op => `<option value=${op.key}>${TBHelpers.htmlEncode(op.text)}</option>`).join('')}
```

eritbh (changes requested):
Hey, thanks so much for taking the time to write a PR for this! I just have a couple suggestions before we can merge it.

dequeues:
Agree with both - I'll update the PR soon. Thanks.

eritbh (approved):

eritbh:
Thanks again for the contribution! This should land in our next feature update.

Issue #550, opened 2022-02-09
Usernote manager, sort by note type.


creesch:
"Requeste" here: https://www.reddit.com/r/toolbox/comments/soiyfs/not_allowing_us_to_sort_usernotes_by_the_type/

Additionally, colors aren't shown in the usernote manager. We might as well add those.

dequeues:
I'll take a look at this in the next few days

creesch:
I think this might already be implemented recently.

creesch:
Ah I was thinking of the filtering PR you did a while ago. https://github.com/toolbox-team/reddit-moderator-toolbox/pull/546 

I think that that PR actually does cover the use case for this issue as well. Although if you also want to add sorting in addition to filtering then that is absolutely fine :)

dequeues:
You're right, it pretty much does cover it. I could add the colors to the usernote though?

creesch:
Sure! :)

eritbh:
Closed in #560

PR #560, opened 2022-03-16
Implement colors for usernote tags

dequeues:
Implement #550

### DIFF
--- a/extension/data/modules/usernotes.js
+++ b/extension/data/modules/usernotes.js
@@ -1008,7 +1008,7 @@ function usernotes () {
                         <div class="tb-un-note-details">
                             <a class="tb-un-notedelete tb-icons tb-icons-negative" data-note="${key}" data-user="${user.name}" href="javascript:;">${TBui.icons.delete}</a>
                             <span class="note">
-                                <span class="note-type">[${color.text}]</span>
+                                <span class="note-type" ${color.key !== 'none' ? `style="color:${color.color}"` : ''}>[${color.text}]</span>
                                 <a class="note-content" href="${val.link}">${val.note}</a>
                             </span>
                             <span>-</span>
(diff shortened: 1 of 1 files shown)


extension/data/modules/usernotes.js
```diff
@@ -1008,7 +1008,7 @@ function usernotes () {
                         <div class="tb-un-note-details">
                             <a class="tb-un-notedelete tb-icons tb-icons-negative" data-note="${key}" data-user="${user.name}" href="javascript:;">${TBui.icons.delete}</a>
                             <span class="note">
-                                <span class="note-type">[${color.text}]</span>
+                                <span class="note-type" ${color.key !== 'none' ? `style="color:${color.color}"` : ''}>[${color.text}]</span>
```
eritbh, line 1011: I know we don't do this in any of the surrounding code right now, but let's HTML encode this value just to be safe:
```suggestion
                                <span class="note-type" ${color.key !== 'none' ? `style="color:${TBHelpers.htmlEncode(color.color)}"` : ''}>[${color.text}]</span>
```

eritbh (changes requested):
Thanks for the PR! Just one suggestion.

extension/data/modules/usernotes.js
dequeues: Ah yes, I totally forgot about that. I'll update it soon

creesch: Good call, though we don't need to worry about injection much here as the subreddit config already has been put through domPurify here: https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbcore.js#L1361

Still, if for some reason we find ourselves with html that is technically safe it could still mess up rendering.

eritbh (approved):

PR #566, opened 2022-04-11
Modnotes support

### LINKED ISSUE #565: Support Reddit-native mod notes
Add some way to view and create native mod notes on old Reddit.

API docs: https://www.reddit.com/dev/api#section_modnote

eritbh:
Fixes #565

### DIFF
--- a/extension/data/init.js
+++ b/extension/data/init.js
@@ -12,6 +12,7 @@ import Config from './modules/config.js';
 import BetterButtons from './modules/betterbuttons.js';
 import DomainTagger from './modules/domaintagger.js';
 import ModMatrix from './modules/modmatrix.js';
+import ModNotes from './modules/modnotes.js';
 import Syntax from './modules/syntax.js';
 import ModButton from './modules/modbutton.js';
 import General from './modules/general.js';
@@ -212,6 +213,7 @@ const coreLoadedPromise = new Promise(resolve => {
         BetterButtons,
         DomainTagger,
         ModMatrix,
+        ModNotes,
         Syntax,
         ModButton,
         General,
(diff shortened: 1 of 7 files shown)


creesch:
Had some quick looks arround and the basis looks good. A few things I noticed (you probably also know or saw): 

- I can't create notes yet for users that have none, clicking the `NN` button (Typo btw?) yields an error in the background. 
- The notes endpoint returns a bunch more information for each "note" and mod action. It appears that for removals and such on a user it includes the thing ID. It might be a nice idea to link to those items. 
- Lot's of API requests. I assume caching is a TODO? I did have a brief look at the finalized API and the batch option might be useful.

eritbh:
"NN" has just been there since before we talked about what to make it, can change later. Yeah, handling notes for users that don't already have any (and also proper handling and display of errors) is TODO.

The only bit of info I don't integrate yet is the thing IDs, everything else should be accounted for. I've put of dealing with thing IDs because I want to be able to link to them, but we can get comment IDs without any information about their post ID which makes it impossible to link to the comment without first making an API requests to find out its post ID, then building the `/comments/postid/-/commentid` URI. If you have a better strategy for that, or if I've missed any other data we could be displaying, let me know.

Caching is a giant TODO at the moment, yes. I put a comment in there somewhere about using the batch API but haven't had a chance to look at it just yet. That will probably be the next thing I work on after implementing note label selection when creating new notes.

creesch:
>  without first making an API requests to find out its post ID, then building the /comments/postid/-/commentid URI. If you have a better strategy for that 

Nah that works for me, it is just when opening the popup. To not delay building of the popup we could explore doing this as an async call where we first build the popup and leave out links where we don't have enough information. Then fire of the api calls to fetch that information and update those entries in the popup once the call is finished. 

We could possibly keep a cache of that as well if we want to.

Issue #570, opened 2022-04-11
chore: fix jsdocs pages thing


creesch:
Just noticed we used travis for that and it has been broken for a while. We probably want to redo it with github actions. 

On PR merge or mast commit:

1. Generate jsdocs https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/package.json#L10
2. Grab the resulting files. 
3. Commit the files here: https://github.com/toolbox-team/source-docs

creesch:
Oh right, I already did this. We still need to see if we can structure functions to actually fall under their modules but other than that it work.

PR #575, opened 2022-05-19
Update repo meta files

eritbh:
Sprucing up our README and CONTRIBUTING documents a bit.

- [x] Remove some old stuff that I don't believe gets used anymore
  - [x] Travis badge, since we don't appear to use it anymore (@creesch might want to confirm everything that needs to be taken off of there is off of there, any secrets or other stuff)
  - [x] IRC badge (this I'm not actually positive about, we might want to keep it around? i'm just not in it anymore)
  - [x] Sauce Labs blurb? do we still use their stuff?
- [x] Reformat files to remove trailing spaces, use consistent header forms, etc.
- [x] Use consistent link style in README (all footnote-style links or all inline links, let's not mix them [except for images I guess])
- [x] Add readme badges for extension versions
  - [x] Chrome
  - [x] Firefox
  - [x] Edge (i forgot)

fixing the rest tomorrow

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,57 +1,59 @@
 # Contributor guidelines
 
-Thinking about contributing to toolbox? Awesome! Here is some information to get you started! 
+Thinking about contributing to toolbox? Awesome! We accept a variety of different types of contributions:
 
-## Reporting bugs 
+## Issues and Ideas
 
-We ask that you first make a post in the [/r/toolbox](https://www.reddit.com/r/toolbox) subreddit before submitting an issue here. This in order to keep clutter out of the issues on github as much as possible. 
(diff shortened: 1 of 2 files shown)


creesch:
Looks good to me! We don't use travis anymore (We need to fix our doc generation to use github actions, I think I made an issue for that) so that is fine. 

> Sauce Labs blurb? do we still use their stuff?

Not really, it is required to use their stuff without paying but we never ended up implementing front-end e2e tests.


> Use consistent link style in README (all footnote-style links or all inline links, let's not mix them [except for images I guess])

I am not sure I follow what you mean but do what you like best and that will be the standard :D

eritbh:
Right, the issue for fixing Travis doc generation was #570. Are we good to remove the IRC link as well or do you still hang out in there? Wanted to get confirmation on that too

creesch:
Oh yeah, irc can go as well.

eritbh:
Cool, this should be all set then!

PR #583, opened 2022-08-18
Use undocumented native endpoints to leave removal reason comments/modmails fixes #582 

### LINKED ISSUE #582: Use undocumented native endpoints to leave removal reason comments/modmails
this issue is written in a really roundabout way because i'm tired and have no energy sorry. but yes this is about "remove as a subreddit"

As pointed out on Discord [here](https://discord.com/channels/535490452066009090/540164483025469490/1009568435803533393), there's an unoffical/undocumented native endpoint for submitting removal reasons. It allows you to send the reason as a comment under a built-in subreddit shared account (this is new and recent). I would guess it will also be used for other things in the future, e.g. to show removal reasons left on users in mod notes or something like that.

The (undocumented) endpoints are `/api/v1/modactions/removal_comment_message` and `/api/v1/modactions/removal_link_message` which take a body like

```jsonc
{
  "item_id": [
    "t1_ikodx0k"
  ],
  "message": "comment goes here",
  "title": "rule title goes here?",
  "type": "public_as_subreddit", // or `public`, `private`, `private_exposed`
  "lock_comment": false
}
```

The endpoints don't appear to require any hard link to the native removal reasons data (the `"title"` field is... something? but it doesn't appear to be reflected anywhere), so we should theoretically be able to modify the current Toolbox removal reasons module to use these endpoints to submit comments/modmails when leaving reasons with arbitrary text. We can also use this to support sending removal reason comments as the subreddit.

creesch:
What the title says :)

### DIFF
--- a/extension/data/modules/removalreasons.js
+++ b/extension/data/modules/removalreasons.js
@@ -52,6 +52,12 @@ export default new Module({
             description: 'Leave removal reasons as a sticky comment.',
         },
         {
+            id: 'reasonCommentAsSubreddit',
+            type: 'boolean',
+            default: false,
+            description: 'Leave removal reason comments with /u/subreddit-ModTeam.',
+        },
+        {
             id: 'actionLock',
             type: 'boolean',
             default: false,
@@ -80,6 +86,7 @@ export default new Module({
     reasonAsSub: reasonAsSubSetting,
     reasonAutoArchive: reasonAutoArchiveSetting,
     reasonSticky: reasonStickySetting,
+    reasonCommentAsSubreddit: reasonCommentAsSubredditSetting,
     actionLock: actionLockSetting,
     actionLockComment: actionLockCommentSetting,
 }) {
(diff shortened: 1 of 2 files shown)


extension/data/modules/removalreasons.js
```diff
@@ -859,37 +874,66 @@ export default new Module({
...
-                        // Lock reply if requested
-                        if (actionLockComment) {
-                            const commentId = response.json.data.things[0].data.id;
-                            self.log(`Fullname of reply: ${commentId}`);
-                            TBApi.lock(commentId).then(() => {
-                                removePopup(popup);
+                        modactionsEndpoint = '/api/v1/modactions/removal_link_message';
+                    }
+
+                    TBApi.apiOauthPOST(modactionsEndpoint, JSON.stringify({
```
eritbh, line 886: I think ideally it'd be good to have these broken out as TBApi helpers, but not a blocker; we can handle that later when this all gets torn down and rebuilt from scratch for #186

eritbh (approved):

extension/data/modules/removalreasons.js
eritbh: oh wow that diff is terrifying when whitespace changes aren't hidden, sorry :V

creesch: Yeah, I did consider making a helper function. But didn't want to actually think too hard about how well thought helper function for this would look like. Also because we don't know exactly yet what we want to use it all for when refactoring. 

And yeah, it scared me as well when I first saw the diff :D

PR #585, opened 2022-08-18
don't resolve init until listeners are added

eritbh:
Follow-up to #584. The usernotes module wasn't adding its jsAPI listeners before allowing its `init` function to resolve; adding some additional `await`s ensures that the listeners will be added before the intializer resolves

### DIFF
--- a/extension/data/modules/usernotes.js
+++ b/extension/data/modules/usernotes.js
@@ -49,23 +49,23 @@ const self = new Module({
             hidden: true,
         },
     ],
-}, function init (initialSettings) {
+}, async function init (initialSettings) {
     startUsernotesManager.call(this, initialSettings);
-    startUsernotes.call(this, initialSettings);
+    await startUsernotes.call(this, initialSettings);
 });
 export default self;
 
-function startUsernotes ({maxChars, showDate}) {
+async function startUsernotes ({maxChars, showDate, onlyshowInhover}) {
     const subs = [],
           $body = $('body');
     const self = this;
     let firstRun = true;
 
-    TBCore.getModSubs().then(() => {
-        self.log('Got mod subs');
-        self.log(window._TBCore.mySubs);
-        run();
-    });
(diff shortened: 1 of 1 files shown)


extension/data/modules/usernotes.js
```diff
@@ -49,23 +49,23 @@ const self = new Module({
...
-}, function init (initialSettings) {
+}, async function init (initialSettings) {
     startUsernotesManager.call(this, initialSettings);
-    startUsernotes.call(this, initialSettings);
+    await startUsernotes.call(this, initialSettings);
 });
 export default self;
 
-function startUsernotes ({maxChars, showDate}) {
+async function startUsernotes ({maxChars, showDate, onlyshowInhover}) {
```
creesch, line 58: `onlyshowInhover` is never actually used as parameter when this function is called?

extension/data/modules/usernotes.js
```diff
@@ -130,8 +121,6 @@ function startUsernotes ({maxChars, showDate}) {
     }
 
     function addTBListener () {
-        const onlyshowInhover = self.get('onlyshowInhover');
```
creesch, line 133: Basically related to the other comment, I am not sure what you moved around here and where it went.

extension/data/modules/usernotes.js
eritbh: The `onlyshowInhover` added here is the value of that setting when the module is initialized; it's part of the `initialSettings` object we get from the `init` function. The `const onlyshowInhover = self.get('onlyshowInhover')` line I removed was fetching the value of the setting a second time, which is unnecessary when we can just grab the value from here.

This was a bit unrelated to the core issue, but refactoring this to avoid an extra asynchronous call did help simplify things.

creesch: ah makes sense :)

creesch (approved):

PR #595, opened 2022-08-23
Modmail sidebar adjustments

### LINKED ISSUE #579: Toolbox buttons break modmail sidebar
A recent change to the user info sidebar in modmail means that the position where Toolbox inserts the user info buttons now results in broken display.

image

The user info no longer scrolls, only the tabs beneath do, so we have limited vertical space to work with. We probably want to reduce our buttons back down to the single-letter labels and insert them underneath the username (`.ModIdCard__UserNameMetaData`) or profile link (`.ModIdCard__UserProfileLink`). We can also steal one of those classes to get text centering for free, to match up with the rest of the display.

### LINKED ISSUE #581: UI improvement options for new Modmail sidebar
As requested around here: https://discord.com/channels/535490452066009090/535490561478492161/1009524587203153991

The new design for the modmail sidebar dedicates way too much space to the user's snoovatar and not enough to the scrollable elements which contain the user's mod notes and conversation history. We could include an option to remove or downsize the snoo and make more space for useful information. I proofed this out with a user style: https://eritbh.me/userstyles/#compact-modmail-user-info

Additionally, the only link to the user's profile page is in the small "view profile" link. We could make the username, display name, or snoo image clickable links to the profile as well.

creesch:
Basically the title. When the old sidebar is still in place it will look the same as it was, with the new sidebar the buttons are moved around. 

image

Fixes #579. Fixes #581.

### DIFF
--- a/extension/data/modules/newmodmailpro.js
+++ b/extension/data/modules/newmodmailpro.js
@@ -114,11 +114,12 @@ export default new Module({
             }
         }
 
+        // A few obivious elements in the sidebar should be clickable and lead to the user profile.
         const $modIdCard = $body.find('.NewInfoBar__idCard .ModIdCard:not(.tb-profile-link-seen)');
         if ($modIdCard.length) {
             $modIdCard.addClass('tb-profile-link-seen');
             const userProfileLink = $body.find('.ModIdCard__UserProfileLink a').attr('href');
-            const wrapElement = `<a href="${userProfileLink}" target="_blank"></a>`;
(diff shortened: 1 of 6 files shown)


extension/data/modules/newmodmailpro.js
```diff
@@ -114,11 +114,12 @@ export default new Module({
             }
         }
 
+        // A few obivious elements in the sidebar should be clickable and lead to the user profile.
```
SpyTec, line 117:
```suggestion
        // A few oblivious elements in the sidebar should be clickable and lead to the user profile.
```

SpyTec (commented):
👀

eritbh (approved):

PR #599, opened 2022-08-24
Update modqueue caching to work with storage api

creesch:
Still some potential for race conditions compared to having a map in memory, but it probably is good enough

It did occur to me that as it is no longer session based it needs to be included in any clear cache mechanism, but I couldn't be bothered to do that now.

As some of our users are subreddit collectors and I didn't want to find out if there is a limit on the amount of storage keys I went with the route of storing all queues in one storage key. 
Which made it slightly more complex, but is probably worth it.

### DIFF
--- a/extension/data/background/handlers/modqueue.js
+++ b/extension/data/background/handlers/modqueue.js
@@ -75,7 +75,7 @@ messageHandlers.set('tb-modqueue', async request => {
     // Instead we wait for the finished event and then reference the cache object.
     if (refreshActive) {
         console.debug('Cache being refreshed, wait for it to be finished');
-        await waitForCacheRefresh(subreddit, thingName);
+        await waitForCacheRefresh(subreddit);
     // The thing timestamp is bigger than the last refresh or cache isn't fresh anymore.
     } else if (thingTimestamp * 1000 > lastRefresh || Date.now() - lastRefresh > 1000 * MODQUEUE_CACHE_TTL) {
         console.debug('Cache needs a refresh');
--- a/extension/data/background/handlers/modqueue.js
+++ b/extension/data/background/handlers/modqueue.js
@@ -1,9 +1,40 @@
 'use strict';
 
 const MODQUEUE_CACHE_TTL = 30;
+const MODQUEUE_CACHE_NAME = 'tb-modqueue-cache';
 
-// Object containing the queue cached data per subreddit
-const queueCache = new Map();
+/**
+ * Sets the modqueue cache for a given subreddit.
+ * @param subreddit string subreddit
+ * @param cacheObject object containing the cache
+ */
+function setQueueCache (subreddit, cacheObject) {
+    return new Promise(resolve => {
+        browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}}).then(result => {
+            result[subreddit] = cacheObject;
+            browser.storage.local.set({
+                [`${MODQUEUE_CACHE_NAME}`]: result,
+            }).then(resolve);
+        });
+    });
+}
+
+/**
+ * Returns the modqueue cache for a given subreddit.
+ * @param subreddit subreddit
+ * @returns {promise<object>}
+ */
+function getQueueCache (subreddit) {
+    return new Promise(resolve => {
+        browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}}).then(result => {
+            if (Object.prototype.hasOwnProperty.call(result[MODQUEUE_CACHE_NAME], subreddit)) {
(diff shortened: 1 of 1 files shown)


extension/data/background/handlers/modqueue.js
```diff
@@ -1,9 +1,40 @@
...
+function setQueueCache (subreddit, cacheObject) {
+    return new Promise(resolve => {
+        browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}}).then(result => {
+            result[subreddit] = cacheObject;
+            browser.storage.local.set({
+                [`${MODQUEUE_CACHE_NAME}`]: result,
+            }).then(resolve);
+        });
+    });
+}
```
eritbh, line 20: Writing this as an async function rather than explicitly returning a `new Promise` makes it a bit simpler to understand at first glance.

```suggestion
async function setQueueCache (subreddit, cacheObject) {
    const result = await browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}});
    result[subreddit] = cacheObject;
    await browser.storage.local.set({
        [MODQUEUE_CACHE_NAME]: result,
    });
}
```

extension/data/background/handlers/modqueue.js
```diff
@@ -1,9 +1,40 @@
...
+    return new Promise(resolve => {
+        browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}}).then(result => {
+            if (Object.prototype.hasOwnProperty.call(result[MODQUEUE_CACHE_NAME], subreddit)) {
+                resolve(result[MODQUEUE_CACHE_NAME][subreddit]);
+            } else {
+                resolve({});
+            }
+        });
+    });
+}
```
eritbh, line 37: Similar to above, could be cleaner as an async function

```suggestion
async function getQueueCache (subreddit) {
    const result = await browser.storage.local.get({[MODQUEUE_CACHE_NAME]: {}});
    if (Object.prototype.hasOwnProperty.call(result[MODQUEUE_CACHE_NAME], subreddit)) {
        return result[MODQUEUE_CACHE_NAME][subreddit];
    }
    return {};
}
```

extension/data/background/handlers/modqueue.js
```diff
@@ -12,69 +43,77 @@ const queueCache = new Map();
...
     // Check if we need to fetch data.
     let lastRefresh = 0;
     let refreshActive = false;
-    const subredditQueueCache = queueCache.get(subreddit);
-    if (subredditQueueCache) {
-        lastRefresh = subredditQueueCache.lastRefresh;
-        refreshActive = subredditQueueCache.refreshActive;
-    }
+    getQueueCache(subreddit).then(subredditQueueCache => {
+        if (Object.keys(subredditQueueCache).length) {
```
eritbh, line 63: Changing this check from `if (subredditQueueCache)` to this keys check seems unnecessary. Can we have `getQueueCache` return `null` or `undefined` by default instead of an empty object and keep this check (and the same type of check done further down) simple?

extension/data/background/handlers/modqueue.js
```diff
@@ -12,69 +43,77 @@ const queueCache = new Map();
...
     const {subreddit, thingName, thingTimestamp} = request;
     // Check if we need to fetch data.
     let lastRefresh = 0;
     let refreshActive = false;
-    const subredditQueueCache = queueCache.get(subreddit);
-    if (subredditQueueCache) {
-        lastRefresh = subredditQueueCache.lastRefresh;
-        refreshActive = subredditQueueCache.refreshActive;
-    }
+    getQueueCache(subreddit).then(subredditQueueCache => {
```
eritbh, line 62: Wrapping this whole section in a callback makes the diff really ugly; lots of whitespace changes masking what's really changing. This call could be refactored to use `async`/`await` instead to avoid the callback and keep the overall structure closer to what we already have.

eritbh (changes requested):

creesch:
Resolved the comments and reworked a bit more to make use of async/await. 

Also fixed two things I did encounter: 
- `refreshActive` being written to an object that wouldn't be read anymore. 
- Edge case where `refreshActive` would remain active if the browser is closed at the right moment (or in the future when the manifest v3 worker times out). `refreshActive` is now also a timestamp and if it is older than the TTL for the cache we can assume something went wrong in a previous call.

eritbh (changes requested):
Just to confirm, are the newly added debug logs left in this PR intentionally?

One more nit to address and then this looks good to me

creesch:
> Just to confirm, are the newly added debug logs left in this PR intentionally?

Yup, should help in the future if I ever need to mess with it as it was now a pain to debug at times.

eritbh (approved):

PR #601, opened 2022-08-24
Clean up unused dev dependencies and build tools

### LINKED ISSUE #597: Clean up package.json (old build dependencies, unused test dependencies, etc)
At this point it is pretty clear that we are not going to do much in the way of unit tests expansion. If we do it likely will be slightly different anyway given the ES6 rewrite. 

In a similar sense there are likely dependencies in there used by `build.js` and therefore no longer relevant as we use rollup.

creesch:
Fixes #597


eritbh:
We also don't use `@rollup/plugin-node-resolve` right now, that can be removed and I'll add it back whenever we get around to using npm for our dependencies

eritbh (approved):

PR #603, opened 2022-08-25
Fix userDetails cache not set and used

creesch:
As the title says, the `userDetails` was never set and also not used. As it is effectively always called on toolbox init the cached modhash and username don't really need to be fetched again. 

Note: PR done in the github web version of vscode so not actually tested yet.

### DIFF
--- a/extension/data/tbapi.js
+++ b/extension/data/tbapi.js
@@ -166,7 +166,7 @@ const userDetailsPromise = (async function fetchUserDetails (tries = 3) {
     try {
         const data = await getJSON('/api/me.json');
         TBStorage.purifyObject(data);
-        TBStorage.setCache('Utils', 'userDetails', data)
+        TBStorage.setCache('Utils', 'userDetails', data);
         return data;
     } catch (error) {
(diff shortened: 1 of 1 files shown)


creesch:
> Note: PR done in the github web version of vscode so not actually tested yet.

Tested and everything works as expected.

eritbh:
Reworked to put the cache lookup in `userDetailsPromise` instead of repeating it in the other functions that rely on `getUserDetails()`. Should be good to go now.

creesch:
This should be ready to merge now: 

- On page load it will always attempt to fetch user details through the API. 
- When succeeded, it now actually does set the cache. 
- The cache is only used as a fallback. For example, when reddit is having backend issues. The reason for only using it as a fallback is that people also switch users and log out, using cached data here would cause issues. 
- The details are kept around in a resolved promise (I was not aware this is a thing, seems like magic but that is mostly on my knowledge gap there) and through there used by the `getModhash`  and `getCurrentUser` functions.

eritbh (approved):

PR #605, opened 2022-08-25
Add spam option to mod macros

### LINKED ISSUE #571: Add a "spam item" option to mod macros
Alongside the "remove item" radio button, would it be possible to add a "spam item" radio button? The intended result would be calling the `remove` API call with the `spam` boolean value set to true when the mod macro is invoked.

If the "spam item" option is set, I'd suggest having the behavior be the same regardless of the "remove item" option state. Just always have the macro remove as spam.

Thanks.

LMBishop:
Closes #571

### DIFF
--- a/extension/data/modules/macros.js
+++ b/extension/data/modules/macros.js
@@ -271,7 +271,7 @@ export default new Module({
             }
 
             if (spam) {
-                actionList += `<br>- This ${kind} will be marked as spam`;
+                actionList += `<br>- This ${kind} will be removed and marked as spam`;
             }
 
             if (approve) {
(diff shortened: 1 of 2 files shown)


extension/data/modules/macros.js
```diff
@@ -269,6 +270,10 @@ export default new Module({
                 actionList += `<br>- This ${kind} will be removed`;
             }
 
+            if (spam) {
+                actionList += `<br>- This ${kind} will be marked as spam`;
```
creesch, line 274: Nitpick, but can you add "and removed" to this? Most people will know,  but it doesn't hurt to be extra clear

creesch:
Had a quick look on my phone, besides one nitpick in wording it looks good to me. I'll finish the review once I get home behind my pc

eritbh:
LGTM after that change, thanks for the PR! Will let @creesch merge

PR #606, opened 2022-08-30
Make more manifest adjustments in rollup per browser.

creesch:
As the title says. The properties added don't break toolbox like `"incognito": "split"` does for firefox. But, they do cause error warnings which also isn't ideal so it is better to remove them for the browsers that throw errors.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -23,9 +23,17 @@ export default ['chrome', 'firefox'].map(platform => ({
                     transform: contents => {
                         const manifest = JSON.parse(contents);
 
-                        // Firefox doesn't support "incognito": "split"
+                        // Firefox doesn't support:
+                        // - "incognito": "split"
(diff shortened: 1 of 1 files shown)


creesch:
Needs some more testing apparantly... When trying this in firefox it seems toolbox gets stuck somewhere. The only log entry is one for tbstorage

image

I am not sure why as the manifest isn't all that different, it is just one property being different. Of course, firefox being firefox we are likely missing a bunch of errors in the console output.

eritbh:
Failure to load in Firefox is due to an error thrown when `version_name` is missing from the manifest. We load the version name into Toolbox's about page directly from the manifest to avoid duplicating it in multiple places, so if it's not causing breaking errors to leave in the unrecognized key, we should probably just keep that as-is.

image

creesch:
Right, I totally forgot about that...

PR #608, opened 2022-09-01
Improvements to modbar module

eritbh:
Pages to make things more bearable, filtering to notes-only or to exclude notes, some other stuff. Notably added `footer` as an option to `TBui.popup()` allowing us to easily set a single universal footer in tabbed popups. Should be in a mergeable state theoretically, but drafting for now so I can avoid thinking which issues this covers.

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -2,7 +2,7 @@ import {Module} from '../tbmodule.js';
 import {getModSubs, link, modsSub} from '../tbcore.js';
 import {escapeHTML, htmlEncode} from '../tbhelpers.js';
 import * as TBApi from '../tbapi.js';
-import {actionButton, drawPosition, icons, popup} from '../tbui.js';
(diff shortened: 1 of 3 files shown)


creesch:
I like the tabbed approach. I'll test it out later today and give a bit more thoughts on it. 

Something else I realized. For modlog items we currently leave out the context. We probably should just make the entries links and have them link to whatever context they refer.

eritbh:
Still working on adding context links, yeah, that was that whole thing about needing to fetch post IDs for comment IDs so it was actually a lot more complicated than it should need to be.

I don't actually see any open issues for the modbar module at all so un-drafting this I guess; should probably start tracking enhancements we need in issues at some point

eritbh:
i am Incredibly Blind, but this PR does neither of those things so my point stands :V

creesch:
Looks good to me. I do think we do want to add a setting for people to choose the default tab the popup opens on. 

But that doesn't need to be in this PR, I can also make an issue so we implement that later.

creesch (approved):

PR #613, opened 2022-09-08
Streamline modbar loading

### LINKED ISSUE #607: Bottom bar and context menu open and close on every page load
The bottom bar and context menu now open up and close every single time a new Reddit page is loaded or reloaded.

Here are two screen recordings showing a page reloading with Chrome and Firefox, respectively:

https://imgur.com/a/2bLJbx9

The bottom bar opens about 5-6 seconds before the end of each video and the context menu opens about a second later. The context menu is usually open for a longer period of time, but I've seen the bottom bar open up for as long as a second too.

I found a previous page load video I created with Toolbox 5.4 and you can't see this happening at all:

https://imgur.com/bcpj0Pt

eritbh:
Fixes #607, a performance issue where a hidden or compact mode modbar would flash the shown state on page load before being hidden. Hiding the modbar on load was previously handled only after populating the "moderated subreddits" list, if enabled, which requires an asynchronous call (`TBCore.getModSubs()`).

This change moves the hiding code up to happen immediately after the modbar and its hidden variant are added to the DOM to avoid the issue, and also changes the moderated subreddits handling to use a callback on `getModSubs()` rather than `await`ing it to avoid blocking other processing (adding event listeners, etc.). I also converted the `getSetting` calls in the module to `getSettingAsync` since I'm in the area and it wasn't a big change.

### DIFF
--- a/extension/data/modules/modbar.js
+++ b/extension/data/modules/modbar.js
@@ -280,6 +280,31 @@ export default new Module({
 `);
 
     $body.append($modBar);
+    $body.append($modbarhid);
+
+    const toggleMenuBar = hidden => {
+        if (hidden) {
+            $modBar.hide();
+            $modbarhid.show();
+            $body.find('.tb-debug-window').hide(); // hide the console, but don't change consoleShowing.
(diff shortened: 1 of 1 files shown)


creesch:
lgm, send it.

PR #615, opened 2022-09-08
Prepare notifications for manifest v3

creesch:
The notification API is idiotic, but that's besides the point. 

- Notification meta data is now stored in a storage key. 
- A stray jquery ajax call has been removed.
- A timeout has been replaced by an alarm (alarms are also no fun at all).

### DIFF
--- a/extension/data/background/handlers/notifications.js
+++ b/extension/data/background/handlers/notifications.js
@@ -1,8 +1,49 @@
 'use strict';
 // Notification stuff
+const NOTIFICATION_STORAGE_KEY = 'tb-notifications-storage';
 
-// We store notification meta data here for later use.
-const notificationData = {};
+/**
+ * Sets the notification ID and meta data object for a given notification.
+ * @param notificationID string notificationID
+ * @param notificationObject object containing the meta data
+ */
+async function setNotificationMetaData (notificationID, notificationObject) {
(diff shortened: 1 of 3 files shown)


extension/data/background/handlers/notifications.js
```diff
@@ -76,14 +118,22 @@ async function sendPageNotification ({title, body, url, modHash, markreadid}) {
...
 // Handle notification creation
 messageHandlers.set('tb-notification', request => {
-    const notificationTimeout = 6000;
     const sendNotification = request.native ? sendNativeNotification : sendPageNotification;
     sendNotification(request.details).then(id => {
-        setTimeout(() => {
-            clearNotification(id);
-        }, notificationTimeout);
+        browser.alarms.create(`tb-notification-${id}`, {
+            delayInMinutes: 1,
```
eritbh, line 135: 1 minute is significantly more than the 6 seconds this was set to before... according to MDN, this accepts a float, so we can do:

```suggestion
            delayInMinutes: 1 / 60 * 6,
```

or alternatively use `when` and specify the absolute time:

```suggestion
            when: Date.now() + 6000,
```

eritbh: this api is dumb of course you can't do that

eritbh (approved):

PR #618, opened 2022-09-09
Use npm for dependency management

eritbh:
Completely removed our `libs` directory in favor of pulling libraries from npm and bundling through Rollup. Generally speaking, I installed the same versions of the libraries in npm as we were already using; some will have to be updated at a later date.

`tbplugins.js` was moved, but git doesn't appear to have picked up on that and just lists it as an added file. The `bindFirst` plugin was removed (it was only used in one place where `.on` was appropriate), the highlight plugin was modified slightly to remove a `with` statement which is disallowed in strict mode and therefore in modules, and `timeago` is now pulled from npm.

The dependency update workflow was removed; it can be replaced by dependabot, which I still need to look into configuring.

CodeMirror CSS is still there for now because I haven't figured out a good way to pull CSS from `node_modules` into the build yet;  this should get its own issue after this is merged.


creesch:
Generally looks good to me :) 

> CodeMirror CSS is still there for now because I haven't figured out a good way to pull CSS from node_modules into the build yet; 

That's fine, eventually we want to migrate to v6 which doesn't have stylesheets like that. 


> the highlight plugin was modified slightly to remove a with statement which is disallowed in strict mode and therefore in modules

This might have angered DeepScan.

eritbh:
Huh, apparently that file wasn't being checked by Deepscan before but now is. The same issue exists in [the original source](https://github.com/tankchintan/highlight-js/blob/master/highlight.js#L60). Fixed by removing the unused variable.

creesch:
> Huh, apparently that file wasn't being checked by Deepscan before but now is. The same issue exists in [the original source](https://github.com/tankchintan/highlight-js/blob/master/highlight.js#L60). Fixed by removing the unused variable.

Oh you know what, it probably was excluded because it is technically third party code and we didn't want to bother with it. I also noticed that there is a bunch of old school var usage in there as well. 

Even things like this beauty

```javascript 
// fix for IE6 suckage
```

Anyway, outside the scope of this PR but probably good to rework some day in the future.

PR #620, opened 2022-09-09
Add timeout in content_script for notifications.

creesch:
Because I was not happy with just relying on the idiotic alarms API....

### DIFF
--- a/extension/data/background/handlers/notifications.js
+++ b/extension/data/background/handlers/notifications.js
@@ -130,14 +130,16 @@ browser.alarms.onAlarm.addListener(alarmInfo => {
 });
 
 // Handle notification creation
-messageHandlers.set('tb-notification', request => {
+messageHandlers.set('tb-notification', async request => {
     const sendNotification = request.native ? sendNativeNotification : sendPageNotification;
-    sendNotification(request.details).then(id => {
-        browser.alarms.create(`tb-notification-${id}`, {
-            delayInMinutes: 1,
-        });
+    const notificationID = await sendNotification(request.details);
+    // The Alarms API is baffling simplistic and limited.
+    // Because of that the ID will be send back to the tab that requested the notification and a timeout used there to relay back to the background worker.
+    // But because it is possible tabs get closed before the timeout we keep the alarm here as a backup.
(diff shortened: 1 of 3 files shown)


extension/data/tbcore.js
```diff
@@ -496,9 +496,10 @@ export const alert = ({message, noteID, showClose}) => new Promise(resolve => {
  * when the notification is clicked
  */
 export async function notification (title, body, path, markreadid = false) {
-    browser.runtime.sendMessage({
+    const notificationTimeout = 6000;
+    const notificationID = await browser.runtime.sendMessage({
         action: 'tb-notification',
-        native: TBStorage.getSetting('GenSettings', 'nativeNotifications', true),
+        native: TBStorage.getSettingAsync('GenSettings', 'nativeNotifications', true),
```
creesch, line 502: Hrm... should this be an await actually?

creesch:
Actually not tested yet, but no eslint issues because I installed code-server on my VPS which can do eslint :P

extension/data/tbcore.js
eritbh: It'd need to be. Promises can't even be serialized in messages iirc, so I don't know what the behavior of this would be without `await`.

creesch: roger roger

extension/data/background/handlers/notifications.js
```diff
@@ -207,12 +209,12 @@ async function onClickNotification (notificationID) {
...
 }
 
-// Handle click/close events on in-page notifications
+// Handle click/clear events on in-page notifications
 messageHandlers.set('tb-page-notification-click', request => {
     onClickNotification(request.id);
 });
 
-messageHandlers.set('tb-page-notification-close', request => {
+messageHandlers.set('tb-page-notification-clear', request => {
```
eritbh, line 217: I don't disagree with this new name, it *is* more accurate, but i'm curious—was there any other reason for the change?

creesch: Before it was triggered when a user closed the application and close made sense. Now the timeout also sends the event and I figured using `clear` is more clear.

eritbh (approved):
probably still test before merge but LGTM

PR #621, opened 2022-09-09
Fix tb buttons not showing sometimes on page load

creesch:
The oldreddit module was a bit too trigger-happy already sending off events before any other modules were able to listen to those events.  This specifically did happen after cache was cleared or when the extension was reloaded. 

Note: The problem is likely a bit more structural but that will be fixed when I rewrite the entire caching mechanism for manifest v3.


eritbh (approved):
Good enough for now

Issue #555, opened 2022-02-15
Manifest v3 support


creesch:
I could have sworn we already had an issue but I can't find it. 

https://developer.chrome.com/docs/extensions/mv3/ 

We can start working on it but only reasonably implement it once firefox is also ready unless we want to keep two versions of the manifest and the affected areas in the code. https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/

Looking at these are some areas we need to make changes to: 

- [x] [Service workers](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview#service-workers) replace background pages. See [also this](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#background-service-workers) and [this page](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/).  It seems like our caching mechanism will be the biggest challenge there as it depends on persistent data, which is no longer possible. 
- [x] Manifest: [Host permissions](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#host-permissions)
- [x] Manifest: [Web-accessible resources](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#web-accessible-resources)
- [x] Double check we are not using any of the [deprecated APIs](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#sunset-deprecated-apis)

The deadline for this appears to be the [start of 2023](https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/).

creesch:
Started looking into this a little bit again. Caching does seem to be the biggest hurdle as far as I can tell. The way our caching is set up it times out after a certain period, we use active timers for that. Once of those timers expires, the background page sends an event to all tabs so cache is cleared across the board. 

This doesn't work with service workers because they are event based and don't stay active. However, it seems like for whatever reason (I honestly can't remember) this event is only handled in TBCore here: https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/tbcore.js#L1431

We also use local storage right now, but that isn't really an issue. 

My current thinking is something along the lines of: 

- We start storing cache together with a timestamp. 
- If a cache value is requested we check against the timestamp if it might be stale. 
- If we do need a proactive timer on caching (needs investigating) we move that to the content_scripts where it basically will trigger based on whatever frontend script first times out after which that script will send a global event to all other tabs. (We already do have this mechanism). 

[...]

PR #632, opened 2022-09-10
setup manifest v3 for chrome

creesch:
Tested and it seems to work, but this should not be merged before #631 and before ratelimiting is rewriten to not use timers (or is removed as reddit clearly doesn't enforce it).


creesch:
image

eritbh (commented):
If we're doing separate Chromium and Firefox manifest files, we should probably get rid of the platform-specific `transform` hook in the rollup config and just include those changes in each file

creesch:
> If we're doing separate Chromium and Firefox manifest files, we should probably get rid of the platform-specific `transform` hook in the rollup config and just include those changes in each file

Yeah true, I'll make the change.

eritbh (approved):

eritbh:
Will this close #555 (the main mv3 tracking issue) or do we plan to keep that open after the fact to address other things?

> this should not be merged... before ratelimiting is rewriten to not use timers (or is removed as reddit clearly doesn't enforce it).

#637 removes ratelimiting entirely and will un-block this PR.

creesch:
> Will this close https://github.com/toolbox-team/reddit-moderator-toolbox/issues/555 (the main mv3 tracking issue) or do we plan to keep that open after the fact to address other things? 

Let's keep it for a little bit

PR #644, opened 2022-09-12
Add modnotes label selector

### LINKED ISSUE #587: Setting labels when creating mod notes
We can currently create notes, but there's no interface for setting the note's label.

eritbh:
Fixes #587. Adds a label select dropdown next to the input for creating a new mod note. Also has some cosmetic changes, tweaking table CSS to look better when the table is narrower than the footer and adding a placeholder to the note text input.

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -239,7 +239,11 @@ function createModNotesPopup ({
                         </option>
                     `)}
                 </select>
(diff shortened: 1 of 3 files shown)


creesch:
Some thoughts: 
- The `Create note` button text is no longer properly aligned.
- The input text area is too small for my taste. 
- Label selection feels a bit awkward. 

Is there a reason you went for something different compared to toolbox usernotes UI? Considering the user base we are targeting it makes sense to me. 

image

PR #653, opened 2022-09-30
Fix for faulty reddit color-scheme implementation

creesch:
Basically reddit implemented color-scheme to `:root` in a very weird way that basically ignores it. Turns out that we didn't explicitly did style some elements text color which then turned white and a few default html elements turned dark. 

This PR makes toolbox's css more explicit and also disables any color-scheme for anything within tb-window popups or overlays.

### DIFF
--- a/extension/data/styles/toolbox.css
+++ b/extension/data/styles/toolbox.css
@@ -150,11 +150,16 @@ body.mod-toolbox-rd.mod-toolbox-extra .tb-input {
     font-size: 1em;
     padding: 5px;
     line-height: 1.4;
+    background-color: #fff;
+    color: #000;
+    color-scheme: none;
 }
 
+body.mod-toolbox-rd.mod-toolbox-extra .tb-input:focus-visible,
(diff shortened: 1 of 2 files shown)


extension/data/styles/toolbox.css
```diff
@@ -150,11 +150,16 @@ body.mod-toolbox-rd.mod-toolbox-extra .tb-input {
     font-size: 1em;
     padding: 5px;
     line-height: 1.4;
+    background-color: #fff;
+    color: #000;
+    color-scheme: none;
 }
 
+body.mod-toolbox-rd.mod-toolbox-extra .tb-input:focus-visible,
```
eritbh, line 158: `:focus-visible` implies `:focus`, so this added selector is redundant and should nothing; if the goal is to get `:focus-visible` behavior then the `:focus` option should be removed

eritbh (changes requested):
thank you reddit very cool

eritbh (approved):

PR #659, opened 2022-10-16
#311 ignore reports

### LINKED ISSUE #311: QT: Ignore reports for selected items
Just like the existing buttons for approving and removing things, should be able to ignore reports on all selected items.

Endercheif:
Closes #311

Adds:
- `tbapi.ignoreReports`
- `ignore reports on selected` button
- `G` access key for i**G**nore reports button


eritbh (approved):
Thanks for the PR! Functionality works well on my build and the code looks good to me.

@creesch do we want the bulk buttons to respect the betterbuttons "ignore reports on approve"/"approve on ignore reports" settings? It doesn't appear that the approve button has any logic for that right now, but I want to be sure/see if that's something we should address in a future change.

creesch:
> do we want the bulk buttons to respect the betterbuttons "ignore reports on approve"/"approve on ignore reports" settings?

I don't really have a strong opinion about that either way. Maybe see if it is something people expect once this future is being used.

PR #664, opened 2022-10-19
feat: options to unban in modmacros

### LINKED ISSUE #529: Add option to unban to modmacros
As the title says, we should make sure that the API doesn't complain if you try to unban users that aren't banned but other than that it shouldn't be too complex. 

https://www.reddit.com/r/toolbox/comments/p7hnnm/feature_request_please_allow_mod_macros_to_unban/

Endercheif:
Closes #529 

Added
- Checkbox to unban in mod macros
- Unban can be done when executing a macro

Changes
- Refactored start of `editMacro`

### DIFF
--- a/extension/data/modules/config.js
+++ b/extension/data/modules/config.js
@@ -262,6 +262,7 @@ const self = new Module({
                         <div class="tb-macro-actions-row">
                             <h2>User</h2>
                             <label><input type="checkbox" id="banuser">ban user</label>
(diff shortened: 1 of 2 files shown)


eritbh (changes requested):

Endercheif:
Fixed. :)

eritbh (approved):
Thanks again!

Issue #486, opened 2021-04-02
Decouple Toolbox dark theme styles from RES nightmode


eritbh:
Asked for [here](https://www.reddit.com/r/toolbox/comments/mi3rnk/i_asked_this_question_once_but_sadly_didnt_get/) and at least one other time that I can't find a link for right now.

Toolbox should provide its own toggle for light/dark theme, rather than relying only on RES's `.res-nightmode` class for it. We can probably still have an option to base toolbox's nightmode on RES's nightmode, but since RES doesn't apply night mode class/styles on new Reddit and there's no good way to get the new Reddit theme from jsAPI, it probably makes the most sense to just add a manual toggle in general settings.

We can probably just replace-all `.res-nightmode` to `.tb-darkmode` in our existing CSS, then add an option that will always add that class to the page, and another option to add that class to the page whenever RES's class is already there.

This will have an impact on subs that use custom styles for Toolbox elements, but if we default the RES sync option to enabled, then there won't be any immediately breaking changes.

creesch:
Having a explicit toggle makes sense to me. However I'd like to approach it a bit differently in regards to this bit you proposed. 

> We can probably just replace-all .res-nightmode to .tb-darkmode in our existing CSS, then add an option that will always add that class to the page, and another option to add that class to the page whenever RES's class is already there.

I rather keep `.res-nightmode` alongside a new class we create for a few reasons: 

- If we don't people currently having RES nightmode enabled with the expectation that it will also turn toolbox dark will suddenly find that this is no longer working. Besides being annoying for them it also annoys us because we have to explain a bunch of times why this doesn't work. 
- RES is much quicker to load and apply the class (due to not having to do a variety of checks) meaning that there never is a bright "flash" because the class is a bit too late to apply.

eritbh:
> people currently having RES nightmode enabled with the expectation that it will also turn toolbox dark will suddenly find that this is no longer working

This is why I proposed we add a second option that's enabled by default that would actively look for the RES nightmode class and add our own dark theme class as well. That way, current nightmode users won't have their workflow broken.

The goal of having our own class would just be to avoid having a bunch of duplicated CSS selectors like `.res-nightmode .stuff, .toolbox-darkmode .stuff`.

> RES is much quicker to load and apply the class (due to not having to do a variety of checks) meaning that there never is a bright "flash" 

[...]

PR #665, opened 2022-10-19
Make toolbox work with reddit dark mode

creesch:
See the title. The other day I noticed that there is an explicit class used by reddit for their dark mode. It even is the same class as they use in modmail. 

However in modmail they add it to the body element and for the rest of reddit they add it to the html element. 

I opted to simply add our class to the html element as well because I didn't feel like adding a bunch of css selectors.

### DIFF
--- a/extension/data/init.js
+++ b/extension/data/init.js
@@ -282,6 +282,7 @@ async function doSettingsUpdates () {
         }
     });
 
+    $('html').addClass('mod-toolbox-rd');
     $body.addClass('mod-toolbox-rd');
     // Bit hacky maybe but allows us more flexibility in specificity.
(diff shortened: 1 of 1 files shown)


creesch:
I guess related to #486 but we still might want to also make it an option.

eritbh:
This has the potential to break some user CSS that assumes `.mod-toolbox-rd` can only exist on `body`, e.g. `.mod-toolbox-rd:not(.some-other-body-class)` will now always match since `html` and `body` don't get the same classes in general. The fix is for people to use `body.mod-toolbox-rd:not(...)` instead.

We don't appear to use this pattern in Toolbox code, and the fix for any stylesheets this does break is pretty simple, so I'm fine with merging this.

eritbh (approved):

PR #680, opened 2022-11-16
User flair in macros

rysie:
This pull request adds mod macro option for assigning a user an existing user flair.

image

image

### DIFF
--- a/extension/data/modules/macros.js
+++ b/extension/data/modules/macros.js
@@ -549,6 +549,12 @@ export default new Module({
                             banReason: `Muted from: ${info.permalink}`,
                         });
                     }
+
+                    if (userflair) {
+                        TBApi.flairUser(info.author, info.subreddit, null, null, userflair).catch(() => {
+                            TBui.textFeedback(`error, failed to flair user (${userflair})`, TBui.FEEDBACK_NEGATIVE);
+                        });
+                    }
                 }
             }
         });
--- a/extension/data/modules/macros.js
+++ b/extension/data/modules/macros.js
@@ -248,6 +248,8 @@ export default new Module({
             ban,
             unban,
             mute,
+            userflair,
+            userflairtext,
             lockthread: lockitem, // saved as lockthread for legacy reasons
             lockreply,
             sticky,
@@ -311,6 +313,10 @@ export default new Module({
 
         if (unban) {
             actionList += '<br>- This user will be unbanned';
+        }
+
+        if (userflair) {
+            actionList += `<br>- This user will be flaired with [ ${userflairtext} ]`;
         }
 
         if (mute) {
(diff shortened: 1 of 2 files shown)


extension/data/modules/macros.js
```diff
@@ -441,6 +447,12 @@ export default new Module({
                         $body.find('.ThreadViewer .InfoBar__control:not(.m-on) .icon-mute').click();
                     }
 
+                    if (userflair) {
+                        TBApi.flairUser(info.author, info.subreddit, null, null, userflair).catch(() => {
+                            status.text(`error, failed to flair user (${userflair})`);
```
eritbh, line 452: `status.text` doesn't exist. It looks like most of our error reporting in this file is done via `TBui.textFeedback('...', TBui.FEEDBACK_NEGATIVE)`, so we might as well use that:

```suggestion
                            TBui.textFeedback(`error, failed to flair user (${userflair})`, TBui.FEEDBACK_NEGATIVE);
```

Error reporting for this module needs to be redone at some point to support reporting multiple errors at the same time anyway, but this will be good enough for now I think.

eritbh (changes requested):
Looks mostly good to me, though I haven't done any local testing yet. I only see one thing that needs to be addressed.

It looks like a lot of this logic could also be reused to enable a similar post flair feature - though that would also require additional logic to not act on comments. We do something similar already for modmail-specific actions, but I think that should probably be its own PR rather than making it a part of this one. @creesch if you think that's a good feature to add I can make an issue for it.

extension/data/modules/macros.js
rysie: Yup you're right, updated the PR with suggested change.

creesch:
Looks good to me as well, but I also haven't verified functionality locally. 

@eritbh I was thinking the same thing about also having it for posts. But doesn't have to be in this PR so an issue to track it seems like a good idea.

eritbh:
> ...so an issue to track it seems like a good idea.

#682 created

extension/data/modules/macros.js
```diff
@@ -441,6 +447,12 @@ export default new Module({
                         $body.find('.ThreadViewer .InfoBar__control:not(.m-on) .icon-mute').click();
                     }
 
+                    if (userflair) {
+                        TBApi.flairUser(info.author, info.subreddit, null, null, userflair).catch(() => {
+                            TBui.textFeedback(`error, failed to flair user (${userflair})`, TBui.FEEDBACK_NEGATIVE);
+                        });
+                    }
```
eritbh, line 454: This logic is inside a `if (TBCore.isNewModmail)` block, which means the flair is only set if the macro is run from modmail.

The structure of the code in this section is a bit weird, having completely different logic for modmail and non-modmail contexts (because for certain actions we want to steal modmail buttons rather than making API requests ourselves) so I think the best way to get this working outside modmail is to just duplicate this code near the other non-modmail user actions, [around line 526](https://github.com/rysie/reddit-moderator-toolbox/blob/4a4aac22ba23c045f2cbb8989f3d32e167b58dfa/extension/data/modules/macros.js#L526-L551).

eritbh (changes requested):
My local test did find one more thing...

extension/data/modules/macros.js
rysie: Thank you! I totally missed that, I focused rather on modmail functionality. Pushed the changes.

eritbh (approved):
All good on my machine! Thanks for the PR~

rysie:
> All good on my machine! Thanks for the PR~

Lovely! Thanks for checking.

PR #687, opened 2022-12-01
Manually rewrite cookies for container requests

### LINKED ISSUE #98: Firefox container storage weirdness
from /u/durinthal on discord:
image

eritbh:
Fixes #98.

Works around Firefox's lack of support for `"incognito": "split"` by manually rewriting the `Cookie` header of outgoing requests to match the tab that initiated the request. This lets Toolbox work properly with Firefox private windows and containers.

This is just a workaround for the request cookie behavior; notably, it doesn't split cache storage on the background page between cookie stores. Having different users logged in to different contexts will result in conflicting information being written back and forth in the cache. That will need more thought, and I think in an ideal world we would split the cache by logged-in user and allow multiple cookie storewith the same logged-in user to share the same cache, rather than naively splitting it based on cookie store. There are probably other aspects of Toolbox that will need more thought for multi-user support, too.

### DIFF
--- a/extension/data/background/handlers/modqueue.js
+++ b/extension/data/background/handlers/modqueue.js
@@ -58,7 +58,7 @@ function waitForCacheRefresh (subreddit) {
     });
 }
 
-messageHandlers.set('tb-modqueue', async request => {
+messageHandlers.set('tb-modqueue', async (request, sender) => {
     const {subreddit, thingName, thingTimestamp} = request;
     // Check if we need to fetch data.
     let lastRefresh = 0;
@@ -100,7 +100,7 @@ messageHandlers.set('tb-modqueue', async request => {
                     limit: 100,
                 },
                 okOnly: true,
-            });
+            }, sender.tab.cookieStoreId);
             const updatedQueue = await response.json();
             const nowRefresh = Date.now();
             const newCacheObject = {
--- a/extension/data/background/handlers/notifications.js
+++ b/extension/data/background/handlers/notifications.js
@@ -65,7 +65,7 @@ function uuidv4 () {
  * Sends a native browser notification.
  * @param {object} options The notification options
  */
-async function sendNativeNotification ({title, body, url, modHash, markreadid}) {
+async function sendNativeNotification ({title, body, url, modHash, markreadid, cookieStoreId}) {
     // If we have the getPermissionLevel function, check if we have permission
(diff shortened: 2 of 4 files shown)


creesch:
>  That will need more thought, and I think in an ideal world we would split the cache by logged-in user and allow multiple cookie stores with the same logged-in user to share the same cache, rather than naively splitting it based only on cookie store. 

Yup, this was also on my mind when I did the work for manifest v3. 
There is a bit of an extra challenge involved with clearing the cache here. For chrome, I just decided on having two caches, on for regular session and one for incognito sessions. That works because we know those are the only two, but gets a bit more complex with per user caching.

creesch:
https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/background/handlers/cache.js#L9 

This is what we have now. Looking at the code in this PR we might be able to leverage the cookieID as a prefix to store cache. Assuming the ID is always the same from related tabs (I assume it is). 

We still need to implement a method that cleans up the cache after a while, but it gets us halfway there.

eritbh:
Now that I'm looking at this again, there are security concerns with this solution. A malicious webpage could make a request to Reddit setting our cookie store header set and use that to send Reddit requests using credentials from other containers. This would require knowledge of the cookie store IDs ahead of time, but I would want to look into ways to mitigate this before releasing.

eritbh:
9d7e8a3 fixes the security concern by storing the `cookieStoreId` associated with a request in a metadata map rather than directly in a header; a randomly generated "request ID" is used to key into the map and retrieve the cookie store ID to use for cookie rewriting. Entries in the map are deleted after use so they can't be reused by other requests.

Because of Chrome mv3 shenanigans, using local state here instead of `localStorage` (like we do for notifications) is *technically* correct, but this logic will only ever be used in Firefox anyway, and even in Chrome it seems incredibly unlikely that the worker will die and be brought back up between initiating an outgoing request and receiving the same request in `webRequest.onBeforeSendHeaders`.

creesch:
Removing this bit of logic should probably also be part of this PR https://github.com/toolbox-team/reddit-moderator-toolbox/blob/master/extension/data/init.js#L145-L149 

At least, I assume it also works with incognito windows.

creesch:
image

Just putting this here as a reference https://discord.com/channels/535490452066009090/535490452066009093/1105190131087982603

PR #745, opened 2023-04-21
Store cache per user based on session cookie.

creesch:
Does as the title says but does require some extensive testing. 

Also still required is a way to clear up caches automatically. Generally speaking it shouldn't be a big problem as most people don't have a ton of accounts. 
I'll probably store a timestamp per usercache as to when it was last interacted with. Then for any user cache interaction we simply also check the timestamps of any other user cache. If that timestamp is over a few hours old we simply clear that cache.

### DIFF
--- a/extension/data/background/handlers/cache.js
+++ b/extension/data/background/handlers/cache.js
@@ -6,7 +6,7 @@ import browser from 'webextension-polyfill';
 
 import {messageHandlers} from '../messageHandling';
 
-const TB_CACHE_PREFIX = browser.extension.inIncognitoContext ? 'TBCache.incognito' : 'TBcache.public';
+const TB_CACHE_PREFIX = 'TBCache';
 
 const storageShortLengthKey = 'Toolbox.Utils.shortLength';
 const storageLongLengthKey = 'Toolbox.Utils.longLength';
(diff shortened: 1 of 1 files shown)


creesch:
Updated the PR and it is now ready for review. Things added in the latest commit: 

- Fallback in case the session cookie is not available for whatever reason. 
- Mechanism to clean user caches that have not been interacted with for 24 hours. (This one needs a proper look at)
- A random fix in webrequests. We also fetch a cookie there and I noticed that the way we do that it never retries in chrome. So I moved the firefox  `firstPartyDomain` thing outside the catch.

eritbh:
> A random fix in webrequests. We also fetch a cookie there and I noticed that the way we do that it never retries in chrome. So I moved the firefox firstPartyDomain thing outside the catch.

I'm not sure this is the right thing to do - `firstPartyDomain` is a Firefox-only option, so retrying with that option *shouldn't* happen in Chrome. The only time `cookies.get()` should throw here is if we're in Firefox and we didn't supply a `firstPartyDomain`, in which case we supply one and call again; if the call doesn't throw but we get an empty result, we can immediately trust that empty result and know there's no cookie there for us to retrieve.

creesch:
> I'm not sure this is the right thing to do

Well, it is what we already did but the way we did it before was not working as intended for sure.  I suppose we could make a browser specific fix in there. The logic when it was first implemented probably was that firefox might not be the only browser supporting that feature at some point...

creesch:
As per the discussion on discord, I was mixing things up with the firstPartyDomain thing and handling. This is now put back the way I found it for `webrequest.js` and implemented in the same way in `cache.js`.

extension/data/background/handlers/cache.js
eritbh, line 1: I don't think it matters much, but since we don't do any cleanup on the old `TBCache.incognito`/`TBCache.public` keys as part of first launch, I don't see any code path that will ever be capable of deleting those unused keys (unless you're logged in as the users `t2_incognito`/`t2_public`, ha).

We don't really have a clean way to handle something like this from first launch anyway, since that's handled on the content script and the background page doesn't expose methods to update arbitrary keys. However, these keys are also never going to be hit by `staleUserCacheCleanup` since they'll never be added to the `userCacheInteractionTimes` localstorage object since `getSessionUserID` will never be called on a tab where the user is logged in as `t2_incognito` or `t2_public`.

I don't think this needs to block release, but it is a bit of a curiosity - it would probably be worth adding some method down the line to allow a user to clear *all* user caches for troubleshooting purposes, and that would at least allow these old keys to be cleaned at that point.

Otherwise this looks good to me and I'm fine merging this unless you think it's worth it to build out some sort of first launch mechanism for this now.

eritbh (approved):

PR #746, opened 2023-04-21
Save context when creating mod notes

eritbh:
Context items are now saved when creating modnotes. Ideally there should be an option not to include this on a per-note basis, but the footer UI is getting pretty cluttered so I'm not including that for now until we can clean that up a bit.


creesch:
> but the footer UI is getting pretty cluttered so I'm not including that for now until we can clean that up a bit.

https://github.com/toolbox-team/reddit-moderator-toolbox/pull/647

;)

creesch (approved):

PR #758, opened 2023-05-24
More consistent relative date display

eritbh:
[Timeago doesn't support operating on elements before they're in the DOM](https://github.com/rmm5t/jquery-timeago/issues/198) because it keeps intervals open to update the element's text and doing that for elements that aren't on the page anymore would leak memory. To work around this, we queue timeago to run after a tick so it happens after the rest of the HTML is on the page.

Interestingly I found that `Promise.resolve().then(...)` works faster than `setTimeout(..., 0)` in Firefox here - the latter results in significant flashing of non-relative date text when opening the popup, whereas the former has an opportunity to act after the target element is added to the DOM but before the next frame is rendered.

<details>
<summary>Screen recordings of both versions</summary>

Here's the `setTimeout(..., 0)` version:

https://github.com/toolbox-team/reddit-moderator-toolbox/assets/4165301/f5627b33-65a8-4ca8-9517-36fcc917c00e

And here's `Promise.resolve().then(...)`:

https://github.com/toolbox-team/reddit-moderator-toolbox/assets/4165301/4a7217c3-66a9-41fc-98e5-d840ca1a7f89

</details>

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -429,9 +429,9 @@ function generateNoteTableRow (note) {
 
     // HACK: timeago only works on elements added to the DOM, so we run it after
     //       a tick, when the caller has added the constructed row to the page
-    setTimeout(() => {
+    Promise.resolve().then(() => {
         $noteRow.find('time').timeago();
-    }, 0);
+    });
 
     return $noteRow;
 }
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -427,7 +427,11 @@ function generateNoteTableRow (note) {
         $noteRow.append('<td>');
     }
 
-    $noteRow.find('time').timeago();
+    // HACK: timeago only works on elements added to the DOM, so we run it after
+    //       a tick, when the caller has added the constructed row to the page
+    setTimeout(() => {
+        $noteRow.find('time').timeago();
+    }, 0);
 
     return $noteRow;
 }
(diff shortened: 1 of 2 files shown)


eritbh:
A friend suggested a slightly less hacky way to do this might be with mutation observers. It's pretty much equivalent to what's going on here, doesn't have the flashing issue, and would work even if we didn't add the popup to the page immediately after creating it. The code is a bit more complicated, though, and it seems unlikely to address problems we'd run into since this is very module-local UI stuff rather than anything more general.

```js
new MutationObserver((mutations, observer) => {
    mutations.forEach(mutation => {
        if (!mutation.addedNodes.includes($noteRow[0])) {
            return;
        }
        $noteRow.find('time').timeago();
        observer.disconnect();
    });
}).observe(document.body, {subtree: true, childList: true});
```

eritbh:
From what I can see most of the existing timeago calls we make aren't really scoped the way I'm doing it here. A lot of the time we just `$('time.timeago').timeago()` which can only apply to things already in the DOM (since `$()` doesn't know about elements that only exist on the stack). I checked through all the other instances where we call `.timeago()` and they also all happen after the relevant elements are added to the DOM - except for one, which handles timestamp display in usernotes (and it looks like it was probably broken by me when I introduced `TBui.pager()` and restructured how notes are rendered there, so it's having the same problem as modnotes).

I'll add a commit to fix the issue there as well, but everything else should be working fine from what I can tell.

creesch (approved):

eritbh:
Existing implementation fixed in the usernotes manager, and relative date handling has also been added to the main user notes popup.

PR #760, opened 2023-05-25
Display context links in modnotes

### LINKED ISSUE #588: View context links in mod notes
The mod notes module doesn't yet provide links to context items - the item that a given mod note was left on.

eritbh:
Fixes #588. Makes note timestamps link to context items on notes.

image

Generating permalink URLs for comments requires fetching the ID of the comment's parent post, and making an API request for every note on a user would be Bad, so a lot of the time I've spent working on this has gone into making a system that allows us to transparently aggregate multiple calls to `/api/info.json` into a single bulk API call to reduce the number of requests we have to make in a short period. The result is the `TBHelpers.createDeferredProcessQueue` function, which collects function calls over a delay period, executes the callback with the list of all collected calls, and then distributes the results of that callback to the original callers through promises. `TBApi.getInfo` is implemented with this helper, grouping requests into larger `TBApi.getInfoBulk` calls and automatically distributing the results to where they're needed. (In the future, other functions that hit `/api/info.json` like `TBCore.getApiThingInfo` can also be changed to go through `TBApi.getInfo`, grouping more requests into bulk calls and giving us some reduction in ratelimit usage for free.)

In addition to this, I think we still need some mechanism to cache parent fullnames, but that kind of cache storage doesn't fit in super well with our current cache model. Parent fullnames don't need to be stored per-user, and each individual item needs its own cache key with its own expiry date, which is hard to do with the current cache system. Additionally, a TTL on the order of days rather than minutes would probably be more appropriate for this info since it's impossible for it to change. I'll need some help coming up with a good way to handle this, I think.

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -356,8 +356,8 @@ function updateModNotesPopup ($popup, {
 
 /**
  * Generates a table of the given notes.
- * @param {object[]} notes An array of note objects
- * @returns {jQuery} The generated table
+ * @param {object} note A note object
+ * @returns {jQuery} The generated table row
  */
 function generateNoteTableRow (note) {
     const createdAt = new Date(note.created_at * 1000);
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -204,15 +204,18 @@ async function getContextURL (note) {
         return null;
     }
 
+    // Split fullname into type and ID
+    const [itemType, itemID] = itemFullname.split('_');
+
     // Post links only require the ID of the post itself, which we have
-    if (itemFullname.startsWith('t3_')) {
-        return link(`/comments/${itemFullname.replace('t3_', '')}`);
+    if (itemType === 't3_') {
+        return link(`/comments/${itemID}`);
     }
 
     // Comment links require the ID of their parent post, which we need to fetch
-    if (itemFullname.startsWith('t1_')) {
+    if (itemType === 't1_') {
         const parentFullname = await getParentFullname(itemFullname);
-        return link(`/comments/${parentFullname.replace('t3_', '')}/_/${itemFullname.replace('t1_', '')}`);
+        return link(`/comments/${parentFullname.replace('t3_', '')}/_/${itemID}`);
     }
 
     // This ID is for some other item type which we can't process
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -153,9 +153,9 @@ function getLatestModNote (subreddit, user) {
 /**
  * Gets the parent fullname of a comment.
  * @param {string} commentFullname Fullname of a comment
- * @returns {Promise<string>}
+ * @returns {Promise<string>} Fullname of the comment's parent post
  */
-async function getParentFullname (fullname) {
+async function getParentFullname (commentFullname) {
     // NOTE: This function isn't called until the module is initialized, so it's
     //       safe to reference `self` in order to read/write settings
     /* eslint-disable no-use-before-define */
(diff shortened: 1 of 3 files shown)


creesch:
> In addition to this, I think we still need some mechanism to cache parent fullnames, but that kind of cache storage doesn't fit in super well with our current cache model. Parent fullnames don't need to be stored per-user, and each individual item needs its own cache key with its own expiry date, which is hard to do with the current cache system. Additionally, a TTL on the order of days rather than minutes would probably be more appropriate for this info since it's impossible for it to change. I'll need some help coming up with a good way to handle this, I think.

How about not basing it of a TTL but rather a length? Use a regular storage key, store an array in it and give it a max length of `N`.  If you want to make it fancy move items that are accessed back to the front so things that are frequently accessed remain "fresh" for longer.

eritbh:
That seems like a decent plan - what do you think a decent array length would be for something like this? Do we do anything similar elsewhere for reference?

creesch:
At first I was thinking 1000 items as that is a reddit default. But I don't see why we can't just do something like 10.000.  
The biggest consideration is actually extension setting storage size limits which is why I suggested a fixed length.

eritbh:
I don't know the specifics of how the storage quota are calculated, but the most common limit seems to be 5MB and a limit of 10k items would put us probably on the order of 100KB? That seems acceptable, as long as there's not too many other places where we might want to apply this in the future... though I guess it'd be pretty easy to adjust the limit down via updates if we need to. I'll roll with that for now

extension/data/modules/modnotes.js
```diff
@@ -147,6 +150,75 @@ function getLatestModNote (subreddit, user) {
...
+    // Post links only require the ID of the post itself, which we have
+    if (itemFullname.startsWith('t3_')) {
+        return link(`/comments/${itemFullname.replace('t3_', '')}`);
+    }
+
+    // Comment links require the ID of their parent post, which we need to fetch
+    if (itemFullname.startsWith('t1_')) {
+        const parentFullname = await getParentFullname(itemFullname);
+        return link(`/comments/${parentFullname.replace('t3_', '')}/_/${itemFullname.replace('t1_', '')}`);
+    }
```
ashen-dawn, line 216:
```suggestion
    const [ typePrefix, id ] = itemFullName.split('_');
    
    switch (typePrefix) {
        // Type 1 is a comment, retrieve the parent post
        case 't1': {
            const parentFullname = await getParentFullname(itemFullname);
            return link(`/comments/${parentFullname.replace('t3_', '')}/_/${id}`);
        }
        
        // Type 3 is a post
        case 't3':
            return link(`/comments/${id}`);
    }
```

ashen-dawn: This just makes it a bit more clear that the code is matching on a type prefix in all cases, rather than coming across on first read as unrelated string matching.

extension/data/modules/modnotes.js
```diff
@@ -147,6 +150,75 @@ function getLatestModNote (subreddit, user) {
...
 
+/**
+ * Gets the parent fullname of a comment.
+ * @param {string} commentFullname Fullname of a comment
+ * @returns {Promise<string>} Fullname of the comment's parent post
+ */
+async function getParentFullname (commentFullname) {
+    // NOTE: This function isn't called until the module is initialized, so it's
+    //       safe to reference `self` in order to read/write settings
+    /* eslint-disable no-use-before-define */
```
eritbh, line 161: I've decided that this is actually terrible and I don't want to do this. The cache interactions should be moved within the intializer scope so we can reference the module as `this` without shenanigans

extension/data/modules/modnotes.js
eritbh: Not going for the `switch` structure, but fbc9f7e implements using `.split('_')` to get the item type and ID separated from each other up front

Issue #771, opened 2023-07-01
Modmail display issues


eritbh:
There have been a couple recent reports of DOM changes in new Reddit making Toolbox stuff stop showing up ([from Discord](https://discord.com/channels/535490452066009090/535490561478492161/1124452631243407371), [from Reddit](https://www.reddit.com/r/toolbox/comments/14ngu2m/all_prewritten_reply_macros_in_modmail_have/)).

eritbh:
Fixed in #773

PR #772, opened 2023-07-01
Fix for #771

adhesivecheese:
Need to change the element we're inserting at, as Reddit changed the names of things.


adhesivecheese:
I *may* have put this on the wrong side; my memory is for shite and I can't remember where it's supposed to go; this places it between the "reply as" dropdown and the reply button.

eritbh:
Closing in favor of #773, which places it on the left of the "Reply as..." dropdown, but thanks for bringing this to our attention!

Issue #777, opened 2023-07-03
Usernotes relative dates are incorrect when the locale date format is not ISO or US


eritbh:
[As reported on Reddit](https://www.reddit.com/r/toolbox/comments/14p7cbw/bug_wrong_timestamps_on_user_notes/)

PR #780, opened 2023-07-04
Timeago helper function

eritbh:
Adds a helper function `TBui.relativeTime()` which creates `<time>` elements that are automatically rendered as relative times, without UI code being required to manually invoke `timeago`. This should help avoid issues like #777 in the future.

[`timeago` needs to be run once the target element is already in the DOM](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/758), so this implementation keeps track of the elements the helper function has created and uses a `MutationObserver` to call `timeago` automatically once they are in the DOM. This is especially helpful for the submission/comment UI builder functions, which until now required consuming code to manually `$result.find('time').timeago()` after the result was added to the DOM.

### DIFF
--- a/extension/data/tbui.js
+++ b/extension/data/tbui.js
@@ -1538,14 +1538,14 @@ export function makeSingleComment (comment, commentOptions = {}) {
 
     // Add submission time
     const $submittedTime = relativeTime(createdAt).addClass('tb-live-timestamp');
-    $buildComment.find('.tb-comment-score').after($submittedTime);
+    $buildComment.find('.tb-comment-score').after(' ', $submittedTime);
 
     // Indicate if item is edited
     if (commentEdited) {
         const editedAt = new Date(commentEdited * 1000);
         const $edited = $('<span class="tb-comment-edited">*last edited </span>')
             .append(relativeTime(editedAt).addClass('tb-live-timestamp'));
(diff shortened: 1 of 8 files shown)


extension/data/tbui.js
```diff
@@ -1845,6 +1850,49 @@ export function pagerForItems ({
...
+/**
+ * Creates a `<time>` element which displays the given date as a relative time
+ * via `$.timeago()`.
+ * @param {Date} date Date and time to display
+ * @returns {jQuery}
+ */
+export function relativeTime (date) {
+    // create element
+    const el = document.createElement('time');
+    el.dateTime = date.toISOString();
```
ashen-dawn, line 1868: So just to make sure we know what's going on here - putting the `dateTime` property on an element allows you to use the `.timeago()` jquery thing?

ashen-dawn (approved):

extension/data/tbui.js
eritbh: Yeah, that's correct. `timeago` is a plugin from npm; it's not directly obvious where we load it from, but you can read more about it here: https://www.npmjs.com/package/timeago

ashen-dawn: Then it looks good

Issue #761, opened 2023-05-26
Typescript


eritbh:
This has been discussed some in the past [on Discord](https://discord.com/channels/535490452066009090/535490452066009093/1031804776725348432), but I'm making an issue for it now to collect previous thoughts because some of the work I've been doing recently would benefit a lot from TS support and it's motivating me to revisit my old proof of concept for this.

Using TS in toolbox would be an incremental thing and would probably follow roughly these steps:

- [ ] Add support for `tsc` as part of our build process (`allowJs` without any code change and make sure the build still functions)
- [ ] Make sure our existing JS doc comments work with type checking (`// @ts-check`, `checkJs`)
- [ ] Start converting files to TS proper
  - Start with files that are basically just free functions with good documented types, e.g. `TBHelpers.js`, `TBApi.js`
- [ ] Figure out what we want to do about JSON types returned from Reddit (do we just make up our own in the return types of TBApi methods? does anyone else maintain a package we can pull in to get types for API responses? do we just give up and type everything as `object`? (please don't just give up and type everything as `object`))
- [ ] Slowly work towards `noImplicitAny` and other stricter parsing options as we work through the rest of the extension
- [ ] at some point replace our old broken docs build process to use typedoc or something

Specific areas where TS would be useful:

- Editing module settings can be greatly improved with TS - the type inference system can be leveraged to tell you what properties are expected on settings of a certain type while you're typing out the object, and it can also be used to make reading/writing a misspelled storage key a build-time error
- Anywhere we're passing around HTML/jQuery objects between functions - it's not always clear what functions expect an HTML string, a jQuery object, or either, even with the documentation we have in place now
- Certain helper functions, particularly those that deal with arrays, would benefit from being described by template types, but jsdoc `@template` is cumbersome and we don't really use it much
- Sharing types between the content script and background page will be easy and helpful for making sure our message passing is sound
- Deepscan has helped with a lot of this already, but TS type narrowing is useful for eliminating certain dead code (e.g. using type information to identify when a branch can never be taken)

creesch:
> old broken docs build process 

Excuse me, but I fixed that at some point :P 

Other than that, sounds like a plan. I'd add a step where we update our contributing docs where we state that we use typescript and link to a getting started resource. Just in the interest of making contributing still as accessible as possible.

PR #783, opened 2023-07-13
TS conversion starting point

eritbh:
Converts the first set of stuff to Typescript for #761. Everything still builds, yay!

- Note the `compilerOptions.outDir` hack in `tsconfig.json` - `@rollup/plugin-typescript` ignores all output options and does its own thing to pull the build output into the rollup bundle, and it will also just let you pass the tsconfig structure directly to the plugin instead of reading from a file, so it's not strictly necessary to have a `tsconfig.json` in the first place. However a lot of tooling relies on having a `tsconfig.json` to know what's going on, so I decided to just point `outDir` to a throwaway directory in `build`.
  - This directory will only actually get used if you manually `tsc` the project, which you should never *need* to do since rollup will choke on TS errors thanks to `noEmitOnError`. Might be useful at some point in the future for debugging more arcane errors or emergent issues in rollup/its TS plugin though.
- There is no linting for TS files because I don't feel like duplicating our entire linter config into `typescript-eslint` rules at the moment - ~~this should get done before this PR gets merged though~~ this will be handled in a separate issue #784 because the ecosystem has changed significantly
  - it would probably make sense for us to break this out into its own shareable config module on npm at some point, so we can share the config easily between this and `toolbox-devvit`
- JS and TS files have different conventions for how file extensions on imports work:
  - In a Javascript file, importing anything *requires* the source file's extension
  - In a Typescript file, importing another Typescript file *requires* that the file extension *not* be present
  - In a Typescript file, importing a Javascript file works whether or not the extension is present

  I've opted to change as little as possible to get the build to work; eventually everything will be Typescript files importing other Typescript files anyway, at which point every import will be extensionless.


tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+	"extends": "@tsconfig/recommended/tsconfig.json",
+	"compilerOptions": {
+		"allowJs": true,
+		// "checkJs": true,
```
ashen-dawn, line 5: I just want to question briefly if this is left here commented out because it's your intention to enable it later, or to signify a default (although I don't believe this was default), or what exactly?  If you want to keep it then a comment describing why would be nice, if not we should maybe pull it to avoid confusion.

ashen-dawn (approved):

tsconfig.json
eritbh: It'll be enabled later per the plan in #761 - I don't have a strong preference whether it should be clarified here or just removed in the meantime

ashen-dawn: I don't have a preference either, I'll leave it up to you.

eritbh: Done in bf5ffaa

eritbh:
this is where the fun begins

PR #785, opened 2023-07-13
Restructure removal reasons popup

### LINKED ISSUE #186: Refactor removal reasons to use UI builders instead of custom popup

eritbh:
Fixes #186. Cleans up the RR window itself to use `TBui.overlay()` and also cleans up some names and other stuff in this file. There's still a lot of work to be done on this, but other stuff that needs to be broken out from here should be informed by what other work we do from here (e.g. whenever we get around to actually integrating native RRs into this popup).

image


eritbh:
re-targeting to v7.x since this is a refactor that doesn't need to go in a patch release

PR #799, opened 2023-08-02
Fix modnotes parent ID caching

### LINKED ISSUE #793: Make modnotes parent fullname cache perform better
https://www.reddit.com/r/toolbox/comments/15acmpq/bug_loading_native_notes_is_painfully_slow_since/
https://discord.com/channels/535490452066009090/535490561478492161/1133859541868630088

don't let it be huge and test the performance with a full cache. mods on big subs are having multiple second slowdowns when opening the window due to cache reads stalling in `purifyObject` for ages

eritbh:
Fixes #793. Reworks the cache of parent fullnames to just be an object in a module-scoped constant, rather than being persistent. This will prevent us from resending the same requests over and over again when users change tabs or pages when looking at the UI, but things will always be re-fetched when the page is reloaded. This is probably good enough in terms of reducing API requests, considering it's not super common to need to reload the page and immediately look at the notes of the same user a second time.

Not storing this persistently means we don't have to structure the cache around what we can save in settings, so we can just use an `Object.create(null)` which will have *much* better lookup performance than a `[string, string][]`. It also cleans up the code significantly, since we don't need to maintain a reference to the module to write to persistent storage.

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -16,9 +16,6 @@ import {
 } from '../tbui.js';
 import TBListener from '../tblistener.js';
 
-/** The maximum number of entries in the parent fullnames cache. */
-const PARENT_FULLNAMES_CACHE_MAX_SIZE = 10000;
-
 /**
  * An object mapping modnote types to human-friendly display names.
  * @constant {object}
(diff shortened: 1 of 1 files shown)


extension/data/modules/modnotes.js
```diff
@@ -151,44 +149,35 @@ function getLatestModNote (subreddit, user) {
...
-
-    // Write result to cache and trim the cache size; fetch the cache value
-    // again before changing it since others may have touched it in the meantime
-    cachedParentFullnames = await self.get('cachedParentFullnames');
-    cachedParentFullnames.unshift([commentFullname, parentFullname]);
-    await self.set('cachedParentFullnames', cachedParentFullnames.slice(0, PARENT_FULLNAMES_CACHE_MAX_SIZE));
-    return parentFullname;
+    // Fetch the parent fullname fresh
+    // Note that we're not awaiting this - we want the full promise
+    const fetched = TBApi.getInfo(commentFullname)
```
ashen-dawn, line 175: Minor nitpick that `fetched` is not entirely accurate here, as it's a not-yet-resolved promise.

ashen-dawn (approved):

extension/data/modules/modnotes.js
eritbh: Good catch, named this before I decided to hang onto the promises themselves instead of the resolved values. Renamed in 30cd83a

PR #805, opened 2023-08-03
Store account creation date as Date object rather than string

### LINKED ISSUE #804: Error loading user data on Toolbox User Profile
The user profile overlay started having issues displaying user profile data:

image

The following error message is generated in the console: `TypeError: date.toISOString is not a function at relativeTime`

image

It appears [this call](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/da7379998b2925af57681b445868781a72c7c323/extension/data/modules/profile.js#L345) to `TBui.relativeTime()` is passing in the argument [as a string](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/da7379998b2925af57681b445868781a72c7c323/extension/data/modules/profile.js#L327) rather than [the expected Date object](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/da7379998b2925af57681b445868781a72c7c323/extension/data/tbui.js#L1870). Because this function is throwing an error, the entire sidebar of the profile fails to be generated.

shiruken:
In order to use the `TBui.relativeTime()` function, the date must be passed as a Date object rather than a string.

Resolves #804


eritbh (approved):

eritbh:
Base change failed because I forgot how PR rebasing works, merge backed-out and new merge into `v6.1.x` branch force-pushed as ea6c12f

Thanks again for the PR~

PR #807, opened 2023-08-04
Style tweaks for controls in window footers

eritbh:
- Controls in window footers are now laid out consistently
  - Text inputs and buttons are styled to be the same height when shown inline with each other in footers
  - Footers are laid out with flexbox with consistent padding and gap between buttons
- Text and other inputs in footers get the same border color as buttons to stand out more against the more saturated footer background
- Select dropdown chevron is now an embedded SVG instead of a PNG - smaller, and looks good on high-density screens or higher zoom levels

Examples of the new styles:

image

image

image


extension/data/styles/tbui.css
```diff
@@ -96,13 +96,40 @@
...
-    height: 36px;
     display: flex;
+    gap: 5px;
+    padding: 5px;
     justify-content: center;
     align-items: center;
     background-color: #CEE3F8;
 }
+/* A form can be the only child of the footer - use it as the flex container */
+.mod-toolbox-rd .tb-window-footer > form {
```
ashen-dawn, line 107: If this rule is specifically for the form being the only child, we may want to do
```suggestion
.mod-toolbox-rd .tb-window-footer > form:only-child {
```

extension/data/styles/tbui.css
```diff
@@ -96,13 +96,40 @@
...
+    display: inherit;
+    gap: inherit;
+    padding: inherit;
+    margin: -5px;
+}
+/* Hide empty text segments to avoid de-centering buttons and stuff */
+.mod-toolbox-rd .tb-window-footer :not(:empty):empty {
```
ashen-dawn, line 114: So correct me if I'm missing something, but `:not(:empty):empty` will select nothing, right?

ashen-dawn (approved):
The styles look good in testing, a few questions about your CSS selectors, but either way I'm approving it.

extension/data/styles/tbui.css
eritbh: i... huh? yeah you're absolutely right, i have no idea how this happened lmao

extension/data/styles/tbui.css
eritbh: I don't think this is really necessary tbh - there's no case where a `form` has
siblings in a footer, and I can't think of a reason we'd need to do that in the
future either. Plus, since these rules set the `gap` and other flex layout stuff
consistently between the `form` and the footer itself, I think this should work
pretty much as-is in the case where there *is* a `form` with arbitrary siblings,
so there's no need to arbitrarily lock this down.

PR #809, opened 2023-08-08
Modnotes: Fetch *all* of them

### LINKED ISSUE #786: Mod notes: Fetch more than one page of notes
We don't actually implement fetching mod notes past the latest 100, so the following doesn't actually fetch *all* notes: https://github.com/toolbox-team/reddit-moderator-toolbox/blob/0c3597f5d8f1d1cf3f7197d0751a85d9623bba60/extension/data/modules/modnotes.js#L589-L597

The mod notes feature has been out long enough now that [older notes are disappearing from that range for people](https://www.reddit.com/r/toolbox/comments/14yni1l/bug_notes_pop_up_says_no_notes_if_the_user_has/). We'll need to implement additional UI and logic for fetching more notes for the user as needed.

This might be the time to stop using a single call to generate all of the tabs, and start using [the `filter` parameter to `/api/mod/notes`](https://www.reddit.com/dev/api#GET_api_mod_notes) to fetch notes for the specific tab being looked at. This will result in more requests if the user switches between tabs, but less if there are a large number of mod actions that have taken place since the last note was added.

eritbh:
Fixes #786. Adds support for fetching all modnotes on a user, not just the latest 100.

[Generator functions (`function*` syntax)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) are used heavily here since they're a very convenient interface for progressively loading pages of API information as we go, without fetching more than the user needs. There are new counterparts to `TBui.pager` and `TBui.pagerForItems` which work using the (asynchronous) iterators that generator functions return: `TBui.progressivePager` and `TBui.progressivePagerForItems`.

Using generators in UI code simplifies the job of loading things as we go a lot. First, we set up a generator function `getAllModNotes` which handles paging through the API and yields every individual note in order. Then, we can make a second generator that iterates over `getAllModNotes()` and passes each note through `buildNoteTableRow`. We then pass that generator into `progressivePagerForItems`, and the pager internals handle loading more items from the generator as needed when the user clicks "load more".

The original `TBui.pager` method has been reimplemented in terms of `progressivePager` since a fair amount of code was shared between the two - it maintains the same API, and the usernotes manager still works fine with it.

Screenshot of the new modnotes UI with the "load more" button displayed:

image

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -154,6 +154,7 @@ function getLatestModNote (subreddit, user) {
  * subreddit that match the given filter.
  * @param {string} subreddit The name of the subreddit
  * @param {string} user The name of the user
+ * @param {string} filter Criteria for filtering notes
  * @returns {AsyncGenerator<any, void, unknown>}
  */
 async function * getAllModNotes (subreddit, user, filter) {
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -398,7 +398,7 @@ function createModNotesPopup ({
 
         // Generate a table for the notes we have and display that
         const $notesPager = progressivePagerForItems({
-            perPage: 10,
+            perPage: 25,
             controlPosition: 'bottom',
             wrapper: `
                 <table class="tb-modnote-table">
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -412,6 +412,9 @@ function createModNotesPopup ({
                     </thead>
                 </table>
             `,
+            emptyContent: `
+                <p>No notes!</p>
+            `,
         }, (async function * () {
             // Yield a table row for each note that belongs in the tab
             for await (const note of tabModNotes) {
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -155,7 +155,7 @@ function getLatestModNote (subreddit, user) {
  * @param {string} subreddit The name of the subreddit
  * @param {string} user The name of the user
  * @param {string} filter Criteria for filtering notes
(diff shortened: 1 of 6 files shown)


ashen-dawn (approved):
It looks good, the generator pattern works well for this and it's all pretty clear what's going on.

eritbh:
Recent commits rework things slightly from before:

- All functions which previously took `(Async)Generator<T, void>` now accept any `(Async)Iterable<T>` and explicitly tease out the `@@asyncIterator`/`@@iterator` to work with if they require working with `.next()` manually
  - this is a fun thing i didn't know about before! generators are both iterable and also their own iterables! very convenient that you can use them either way. however it was not at all obvious at first and i spent a while being confused about this and about which interface was the most general
- Pulled in the `iter-ops` library to significantly clean up composition of iterators/generator functions
- Because `iter-ops` trivialized the implementation of `progressivePagerForItems` to such a degree, removed that function in favor if implementing the paging behavior directly in `createModNotesPopup()`
  - reasoning behind this is summarized in the message of b03b785:
    > The wrapper argument makes this function janky and hard to reason about, and using iter-ops pipes makes it pretty trivial to implement the logic here from consuming code, while maintaining finer control over how the items are placed in the DOM. The modnotes popup has been reimplemented to make proper use of a <tbody> thanks to this extra control.

Diff went from +375/-141 to +336/-132; overall I think these changes mildly decrease the complexity introduced in this branch.

ashen-dawn:
Okay we need to take the car in for some work but I'm leaving the tab open so I hopefully remember to re-review it when we get back.

ashen-dawn (approved):
`iter-ops` really cleans this up.  Definitely a bit more abstract but still very readable, it looks good to me.

PR #810, opened 2023-08-09
Typescript conversion: TBLog, TBConstants

eritbh:
Converts `tblog.js` and `tbconstants.js` to Typescript. Having the logger done opens up some more things, because the inferred types prior to this were so bad that using the logger in any capacity from TS files required `@ts-expect-error` on every single call.

There was some other cleanup I wanted to do on the logger to make it slightly cleaner in implementation, but I decided that if I keep trying to get it perfect I'm never actually going to push anything, so I've just elected to add naive types to the existing code and call it a day.

#761


extension/data/tblog.ts
```diff
@@ -44,7 +46,9 @@ function log (caller, type, ...args) {
         callerName = caller;
     }
     // Get the appropriate styles for this log type, and send the message
-    const {color, background, text, func} = logTypes[type];
+    const config = logTypes[type];
+    const {color, background, func} = config;
+    const text = ('text' in config) ? config.text : type;
     func(
         // First part of the message line
         `tb: %c[${callerName}] %c${text || type}`,
```
ashen-dawn, line 54: This `text || type` is redundant as you've checked for `'text' in config` above.

extension/data/tblog.ts
```diff
@@ -57,30 +61,33 @@ function log (caller, type, ...args) {
...
  */
-function TBLog (caller) {
+function TBLog (caller?: string | {id: string}) {
     // Create a new object
-    const obj = {};
+    const obj: Partial<Logger> = {};
     // The object gets a function for every log type
-    for (const type of Object.keys(logTypes)) {
+    for (const type of Object.keys(logTypes) as LogType[]) {
         // `this` arg is the caller, first arg is the log type
```
ashen-dawn, line 80: I know you didn't change this line, but this comment is misleading, as the `this` arg is actually `undefined`, and the caller is being given as the first parameter.

ashen-dawn (approved):
A few minor nitpicks, but it looks rather good.

extension/data/tblog.ts
eritbh:
```suggestion
        `tb: %c[${callerName}] %c${text}`,
```

extension/data/tblog.ts
eritbh:
```suggestion
        // `this` arg is not provided
```

ashen-dawn (approved):
Looks good

PR #811, opened 2023-08-10
Use dprint for formatting

### LINKED ISSUE #784: Rework project formatting
We're moving to Typescript (#761) and [`eslint` and `eslint-typescript` no longer recommend using themselves as formatting tools](https://typescript-eslint.io/linting/troubleshooting/formatting). The TS PR #783 includes no formatting for Typescript files and we're gonna have to move our project's formatting config to something else (probably `prettier`, maybe `dprint`). This will probably involve reformatting the entire codebase under a new standard, because this repo's existing conventions are different from exosystem standards and I don't really feel like writing a whole new config for a new tool.

Any PR that reformats the project should add a `.git-blame-ignore-revs` file for the commit that actually applies the new formatting, and that PR should *not* be squash-merged to make sure that the commit referenced there maintains the correct SHA and doesn't contain non-automated changes.

eritbh:
Fixes #784. Sets us up to use `dprint` for formatting the project. Removes formatting-specific eslint rules.

I lean towards using `dprint` isntead of `prettier` for formatting because it's more configurable and lets us maintain a style relatively similar to what we already have. This PR touches lots of files, but the only things that are really super different are line wrapping behaviors in certain areas and disallowing declaring multiple variables in a single `let`/`const` statement. It also has much more sane formatting for complex multiline ternary expressions than `prettier`.

`dprint` uses the same set of rules for formatting both Javascript and Typescript, and also does formatting for JSON and Markdown files.

I'll need to re-enable non-squash merges for this repo before I merge this - the PR adds a `.git-blame-ignore-revs` file with commit hashes of specific commits in this PR, so those commits need to not be squashed when merging for it to work correctly.


dprint.json
```diff
@@ -0,0 +1,34 @@
...
+        "functionExpression.spaceBeforeParentheses": true,
+        "getAccessor.spaceBeforeParentheses": true,
+        "setAccessor.spaceBeforeParentheses": true,
+        "method.spaceBeforeParentheses": true
+    },
+    "json": {
+        "indentWidth": 4
+    },
+    "markdown": {},
+    "excludes": [
```
ashen-dawn, line 25: I'm going to suggest we add the `./build` directory to this as well, as people might want to run the format command on their local repo before pushing (or this might in the future be integrated as a pre-commit hook), and it can currently fail on the following files if you've run a local build:
```
./build/firefox/manifest.json:
./build/chrome/manifest.json:
./build/firefox/data/background/index.js:
./build/chrome/data/background/index.js:
./build/chrome/data/init.js:
./build/firefox/data/init.js:
```

ashen-dawn (changes requested):

dprint.json
eritbh: Huh, not sure how I didn't find this myself - good catch

ashen-dawn: Ideally it would just have an option for "ignore gitignored files" but it may not have that.

ashen-dawn: Yeah looks like they don't currently support just pulling in the gitignored file
(see [here](https://github.com/dprint/dprint/issues/378))

eritbh: aa39fbb

ashen-dawn (approved):
It never fails on build files, looks good.

PR #820, opened 2023-08-23
Handle null or undefined last backup

### LINKED ISSUE #819: "Last Backup" shows days since January 1, 1970
If you have not ever backed up your settings, the plugin is currently misidentifying this as a backup taken at the unix epoch:

image

ashen-dawn:
Fixes #819.  Upon investigating I found that `lastExport` was getting assigned `null`.  Adding an exception to the `TBStorage` function seemed inappropriate, so a fallback in the case of `null` or `undefined` seems like the best way to solve it.  As you can see below it does restore the intended functionality of showing "Never" in this case

image


eritbh (approved):
Looks good, thanks for the PR~

PR #884, opened 2024-01-25
Reduxify `TBui.textFeedback`

eritbh:
We use Redux now!

`TBui.textFeedback` rendering is now handled via a persistent "container" component which sits in `AppRoot` forever and reads the current text feedback state from the global Redux store. The actual `TBui.textFeedback` function takes the same arguments it did before and works functionally the same, but rather than the DOM manipulation being handled in the function body, it just shells out to a Redux thunk which handles adding and removing the feedback message from Redux state after the appropriate timeout. The container element in the app root then sees these changes in state and handles actually rendering the message.

The fade-out animation doesn't exist now because animations in React are hard but I'll get to that eventually. we're gonna need to pull in another animations lib for that to replace our usage of jQuery effects (`$.fadeOut()` et. al)

### DIFF
--- a/extension/data/AppRoot.tsx
+++ b/extension/data/AppRoot.tsx
@@ -1,9 +1,17 @@
+import {Provider} from 'react-redux';
+
+import store from './store';
+
 import {PageNotificationContainer} from './components/PageNotificationContainer';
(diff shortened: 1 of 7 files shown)


ashen-dawn (approved):

eritbh:
fuck it we ball

PR #892, opened 2024-02-22
Fix broken modnote hyperlinks

### LINKED ISSUE #890: modnote module links are broken on comments with parent comments
To reproduce this bug:

* Enable beta features in the Toolbox settings
* Create a modnote on a comment with a parent
* Click on the link that was created

toolbox_broken_link

You get a page not found error instead of being linked to the comment.

The reason for this is on [line 213 of modnotes.js within the getParentFullname function](https://github.com/toolbox-team/reddit-moderator-toolbox/blob/e542efd60ce914a65a065b103c80060e4a4718ed/extension/data/modules/modnotes.js#L213). The `parent_id` value will be different depending on whether or not the noted comment is top-level or not. If it is a top-level comment then this will be the submission ID. If it is not a top-level comment then this will be the comment ID of the parent comment.

This results in a malformed link that looks like this:

    https://www.reddit.com/comments/{parentCommentID}/_/{commentID}

Rather than the expected:

   https://www.reddit.com/comments/{parentSubmissionID}/_/{commentID}

The fix is simple, just replace `info.data.parent_id` with `info.link_id`. This will always result in the correct URL as far as I can tell

jenbanim:
Fixes #890 

I really hope I did this right lol

### DIFF
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -210,7 +210,7 @@ export function getParentFullname (commentFullname) {
     // Fetch the parent fullname fresh
     // Note that we're not awaiting this - we want the full promise
     const parentFullnamePromise = TBApi.getInfo(commentFullname)
-        .then(info => info.data.parent_id);
+        .then(info => info.data.link_id);
 
     // Write to cache and return
     parentFullnamesCache[commentFullname] = parentFullnamePromise;
--- a/extension/data/modules/modnotes.js
+++ b/extension/data/modules/modnotes.js
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
 }
 
 /**
- * In-page cache of comment fullnames to the fullnames of their parent posts.
+ * In-page cache of comment fullnames to the link ID of their submission.
  * Values of this object are promises which resolve to fullnames, rather than
  * bare strings - we keep the promises around after they're resolved, and always
  * deal with this cache asynchronously.
  * @constant {Record<string, Promise<string>>}
  */
-const parentFullnamesCache = Object.create(null);
+const linkIDCache = Object.create(null);
 
 /**
- * Gets the parent fullname of a comment.
+ * Gets the link ID of a comment's corresponding submission.
  * @param {string} commentFullname Fullname of a comment
- * @returns {Promise<string>} Fullname of the comment's parent post
+ * @returns {Promise<string>} ID of the comment's submission
  */
-export function getParentFullname (commentFullname) {
+export function getLinkID (commentFullname) {
     // If it's in cache, return that
-    const cached = parentFullnamesCache[commentFullname];
+    const cached = linkIDCache[commentFullname];
     if (cached) {
         return cached;
     }
 
-    // Fetch the parent fullname fresh
+    // Fetch the link ID fresh
     // Note that we're not awaiting this - we want the full promise
-    const parentFullnamePromise = TBApi.getInfo(commentFullname)
+    const linkIDPromise = TBApi.getInfo(commentFullname)
         .then(info => info.data.link_id);
 
     // Write to cache and return
-    parentFullnamesCache[commentFullname] = parentFullnamePromise;
-    return parentFullnamePromise;
+    linkIDCache[commentFullname] = linkIDPromise;
+    return linkIDPromise;
 }
 
 /**
@@ -239,10 +239,10 @@ async function getContextURL (note) {
         return link(`/comments/${itemID}`);
     }
 
-    // Comment links require the ID of their parent post, which we need to fetch
+    // Comment links require the link ID of their submission, which we need to fetch
     if (itemType === 't1') {
-        const parentFullname = await getParentFullname(itemFullname);
-        return link(`/comments/${parentFullname.replace('t3_', '')}/_/${itemID}`);
+        const linkID = await getLinkID(itemFullname);
+        return link(`/comments/${linkID.replace('t3_', '')}/_/${itemID}`);
     }
 
     // This ID is for some other item type which we can't process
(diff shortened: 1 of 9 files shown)


extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
 }
 
 /**
- * In-page cache of comment fullnames to the fullnames of their parent posts.
+ * In-page cache of comment fullnames to the link ID of their submission.
```
eritbh, line 190:
```suggestion
 * In-page cache of comment fullnames to the fullnames of their submissions.
```

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
  * Values of this object are promises which resolve to fullnames, rather than
  * bare strings - we keep the promises around after they're resolved, and always
  * deal with this cache asynchronously.
  * @constant {Record<string, Promise<string>>}
  */
-const parentFullnamesCache = Object.create(null);
+const linkIDCache = Object.create(null);
```
eritbh, line 196: `submissionFullnamesCache`

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
 
 /**
- * Gets the parent fullname of a comment.
+ * Gets the link ID of a comment's corresponding submission.
```
eritbh, line 199:
```suggestion
 * Gets the fullname of a comment's corresponding submission.
```

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
  * @param {string} commentFullname Fullname of a comment
- * @returns {Promise<string>} Fullname of the comment's parent post
+ * @returns {Promise<string>} ID of the comment's submission
```
eritbh, line 201:
```suggestion
 * @returns {Promise<string>} Fullname of the comment's submission
```

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
  */
-export function getParentFullname (commentFullname) {
+export function getLinkID (commentFullname) {
```
eritbh, line 203: Suggest renaming to `getSubmissionFullname`

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
     // If it's in cache, return that
-    const cached = parentFullnamesCache[commentFullname];
+    const cached = linkIDCache[commentFullname];
     if (cached) {
         return cached;
     }
 
-    // Fetch the parent fullname fresh
+    // Fetch the link ID fresh
```
eritbh, line 210:
```suggestion
    // Fetch the submission fullname fresh
```

extension/data/modules/modnotes.js
```diff
@@ -187,34 +187,34 @@ async function* getAllModNotes (subreddit, user, filter) {
...
     // Note that we're not awaiting this - we want the full promise
-    const parentFullnamePromise = TBApi.getInfo(commentFullname)
-        .then(info => info.data.parent_id);
+    const linkIDPromise = TBApi.getInfo(commentFullname)
```
eritbh, line 212: Suggest renaming to `submissionFullnamePromise`

extension/data/modules/modnotes.js
```diff
@@ -239,10 +239,10 @@ async function getContextURL (note) {
         return link(`/comments/${itemID}`);
     }
 
-    // Comment links require the ID of their parent post, which we need to fetch
+    // Comment links require the link ID of their submission, which we need to fetch
     if (itemType === 't1') {
-        const parentFullname = await getParentFullname(itemFullname);
-        return link(`/comments/${parentFullname.replace('t3_', '')}/_/${itemID}`);
+        const linkID = await getLinkID(itemFullname);
```
eritbh, line 244: Suggest renaming to `submissionFullname` - this code only cares about the unprefixed ID, so the comment above is still correct, but this variable does store the prefixed form and we don't do do the prefix stripping until the next line

eritbh (changes requested):
Hey, thanks so much for submitting a PR for this! Really appreciate upstreaming fixes like this.

You were definitely right to update names and comments for this change, but there's some specific terminology we use in the project that I'd like to use here to disambiguate things a little more:

- Let's use "submission" everywhere rather than "link". "Link" sometimes means "any submission" but other times means "a submission which is not a self/text post," so we avoid using it in code that's relevant to any submission.
- "Fullname" is used to disambiguate thing IDs which also include the thing type prefix (`t1_`, `t3_`, etc) - whereas "ID" can also refer to the form without the prefix, as used in post/comment URLs and also certain API calls. (Reddit itself is even inconsistent about this - not every `*_id` field returned by reddit's various APIs is a fullname...) So any variable which works with prefixed IDs should refer to them as "fullnames" rather than just "IDs" to signal that the prefix is present and may need to be stripped later.

I've left comments about specific changes I'd recommend - feel free to make these updates yourself, or just enable the "Allow maintainer changes" option at the bottom of the PR sidebar in Github and I'll tweak them myself and get this merged. Sorry for the nitpicks!

jenbanim:
No worries about nitpicking, I totally understand the value of consistent naming. I wish Reddit itself were this consistent (eg. "pinned" vs "stickied" posts). Unfortunately I wasn't able to hit the "allow maintainer changes" button since that feature [seems to be broken when you fork to an organization](https://github.com/orgs/community/discussions/5634) as I did, ah well, lessons learned

Thanks for bearing with me, I think this should be good to go now

eritbh (approved):
Thanks again!

PR #896, opened 2024-02-28
Detect logged-in users on sh.reddit.com

LewsTherinTelescope:
Toolbox short-circuits if it doesn't find a logged-in user, which means that even functionality that _does_ work out-of-the-box on sh.reddit—namely the bottom bar—will never run. This does NOT fix anything actually broken by the redesign, but it's better than nothing.

### DIFF
--- a/extension/data/init.ts
+++ b/extension/data/init.ts
@@ -110,10 +110,18 @@ async function checkReset () {
  * @returns {Promise<void>}
  */
 async function checkLoadConditions (tries = 3) {
+    let loggedInShreddit = false;
     let loggedinRedesign = false;
     let loggedinOld = false;
 
     const $body = $('body');
+
+    // Check for re-redesign (sh.reddit.com)
+    if (
+        $body.find("shreddit-app[user-logged-in=true]")
+    ) {
+        loggedInShreddit = true;
+    }
 
     // Check for redesign
     if (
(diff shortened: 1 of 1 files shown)


eritbh:
Base changed to the `v7.x` branch since that's where new feature development is happening until I cut a final patch release with the stuff currently on `master`.

Modules generally assume that if they get run, then toolbox must be running in either old Reddit, new Reddit, or new modmail. As a result, merging this as it is now has the potential to introduce issues where modules make incorrect assumptions about what platform they're on - logic like "if this isn't new reddit, and it's not new modmail, then we must be on old Reddit" is extremely common. For that reason I'm going to hold off merging this until I do some work to improve the platform detection logic used across Toolbox modules, to ensure things that expect to run against a specific frontend don't run unless we can positively identify being loaded on that frontend.

This is a good strategy for handling loading of core stuff though, and it'll be merged down the line as part of v7. Thanks for the PR! Feel free to hit me up in our Discord server if you're curious about other stuff that's going on in this v7 branch.

eritbh:
I've opened #916 which refactors the platform detection logic out of `init.ts` into its own file, `util/platform.ts`. I took the liberty of integrating this PR's contents into the new one in [6ed5477](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/916/commits/6ed54778b953f893b3e88f57fde4395cefeec977), listing you as a co-author. Since this has made it over there I'm going to close this PR for now; feel to subscribe to the new one to be notified when your code lands in master. Thanks again for the contribution!

PR #914, opened 2024-03-22
Fix logic for adding the modnotes window to the DOM

eritbh:
The modnotes window was previously inserted in the DOM as a direct child of `ModNotesUserRoot` which resulted in it being a descendent of the `.tagline` element on old Reddit, which created style issues (unwanted underlines in buttons when hovered, etc). This was done because positioning the popup next to the badge in the DOM made it very easy to get the window to appear close to the location of the badge when first opened.

This change properly recreates the behavior of other windows, where the window is a child of `document.body` via a React portal, and has its initial position on screen set based on the screen coordinates of the click event that causes it to be shown (via the `TBui.drawPosition` helper). This avoids the unintended style issues and also ensures that the window appears precisely where it did before it was converted to React.


ashen-dawn (approved):

eritbh:
https://discord.com/channels/535490452066009090/535490452066009093/1220738364895137823

eritbh — Today at 10:18
this code does actually introduce an issue i forgot to mention in the PR description, which is that because the window has an <input autoFocus/> at the bottom of it, the page scrolls around to ensure that input is on screen when the window is displayed
so i do still need to mess with it a bit to get that to not happen
ashen-dawn — Today at 10:19
Would it be worth considering taking the autoFocus off and then focusing it imperatively with JS after it's rendered?
That way you can make sure it's on screen first
eritbh — Today at 10:19
yeah i was considering something like that, just taking a ref and manually focusing it in requestAnimationFrame or something like that
ashen-dawn — Today at 10:19
Yeah set the window position, then run a requestAnimationFrame callback to make sure it gets painted there, then focus it
That would work

eritbh:
dba87c1 should fix the scrolling-on-focus issues

PR #916, opened 2024-03-22
Refactor platform/login detection and add *extremely* basic Shreddit support

eritbh:
Refactors platform-specific login detection out of `init.ts` and into its own set of utilities in `util/platform.ts`. Includes strategies for dealing with Shreddit, which means that this PR enables some basics of Toolbox to function under Shreddit.

This PR is *not* going to deal with fixing issues in Shreddit, and there are some extremely obvious ones; once this is merged I'll start filing bugs for functionality that's missing or broken on that platform. This PR is focused solely on refactoring platform detection logic.


eritbh:
Example of some extremely broken Shreddit styles which are going to stay broken for now until I commit the style isolation work I have locally: 
image

Seems like global things like personal notes and the settings/sub config windows are *functionally* fine though, so I'm going to merge this as-is and get it into the next beta build

PR #917, opened 2024-03-25
Bringing back beta builds, baby

### LINKED ISSUE #802: Move modnotes out of beta
Depends on #786, #801

eritbh:
couldn't help myself

## Meta changes

- The way we handle version strings in the manifest is kind of different now
  - Chrome and FF both support up to 4 version segments; we have only used 3 until now, but in order to facilitate iterating multiple beta releases per stable version I'm going to be using the fourth segment as a prerelease build indicator from now on
  - I changed up the version parsing code in tbcore a fair bit to facilitate this
  - [x] todo: write some new documentation about how this is supposed to work. it's kind of confusing
- Adds a new deploy workflow for publishing beta builds to two new extension listings
  - Builds for release listings will now explicitly not be created for tags matching `v*-*` - i.e. release builds will be created for tags like `v7.0.0` but not `v7.0.0-beta.1`
  - Builds for the beta channels will be created on all `v*` tags (including non-beta tags)
- Added support for new build environment variables used primarily for beta builds
  - `BUILD_TYPE` - can be `dev`, `beta`, or `stable` - defaults to `dev`
  - `BUILD_SHA` - used for CI builds to provide the commit hash the build came from
    - `dev` builds aren't expected to have this since they're built from local changes, but `beta` and `stable` builds are meant to be distributable which means they will shout at you if you don't provide this in the build. these builds should always be handled by CI anyway though so don't worry about that too much

## Functional changes

- the "beta mode" setting in toolbox is now gone. anything that was switched on that before is now switched on the current build being a beta build
  - the modnotes module and the queuetools "group by parent submission" setting are both taken out of beta. so this PR fixes #802 
- `dev` and `beta` builds now have a version string indicator on the modbar; click it to copy the version string to the clipboard
    image
  - this format is also now used as the version string in the "toolbox info" table in submissions to /r/toolbox
    image
  - similar detail added to version string in the about page on non-stable builds:
    image
- as a somewhat-related quality of life thing for myself specifically, dev builds of toolbox now skip displaying news notes entirely
- removed the unused `build:release` script since all our releases are done through CI anyway. manually running `node release.mjs` is still useful for creating the version tag commits that kick off the CI builds though so that stays around


.github/workflows/publish-beta.yml
```diff
@@ -0,0 +1,91 @@
...
+    - "v*"
+
+jobs:
+  build-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16'
```
ashen-dawn, line 14: this is like . . . _very_ out of date, have we considered upgrading?

CONTRIBUTING.md
```diff
@@ -77,3 +81,33 @@ Once you've built the extension, you can load it up in your browser for testing:
...
+| Manifest `version` | Git tag         | Release type |
+| ------------------ | --------------- | ------------ |
+| `6.1.13.1`         | `v6.1.13`       | stable       |
+| `7.0.0.1`          | `v7.0.0-beta.1` | beta         |
+| `7.0.0.2`          | `v7.0.0-beta.2` | beta         |
+| `7.0.0.3`          | `v7.0.0`        | stable       |
+
+### Tagging a New Release
+
+1. Make sure the working directory is clear and you're on the `master` branch.
```
ashen-dawn, line 104: can the `release.mjs` script be made to check for this and fail if it's not?

extension/data/tbcore.js
```diff
@@ -605,9 +626,14 @@ async function fetchNewsNotes (sub) {
 
 /** Fetch and display all news notes. */
 export function displayNotes () {
+    // dev releases skip all notes for my own sanity
```
ashen-dawn, line 629: smart

ashen-dawn (approved):
Looks good, makes sense

.github/workflows/publish-beta.yml
eritbh: was gonna do this separately but decided to just do it now anyway. [81ecff4](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/917/commits/81ecff4fc7057235769647b4fa6254e11058fcac)

CONTRIBUTING.md
eritbh: all this and more in [731b18d](https://github.com/toolbox-team/reddit-moderator-toolbox/pull/917/commits/731b18dd26b71a3acd58c004affa83234903a229)

PR #1020, opened 2024-12-10
React 19

eritbh:
Upgrades to React 19 and bumps related dependencies to versions that are compatible with the new major React version. Tweaks some code to account for new recommended patterns from the [release notes](https://react.dev/blog/2024/12/05/react-19) and [migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide).

`react-redux` and `@reduxjs/toolkit` aren't yet updated to React 19, but PRs are in progress. Installing dependencies without these being upgraded requires `--force` to ignore the mismatched versions of their peer dependency on React, so this fails CI until those new versions are released.

Supersedes #1015 (#1023), #1017, #1018, #1019 (#1021), #1022.


eritbh:
New Deepscan issues are triggered by rules which haven't been updated for React 19 - `ref` is in fact a normal prop now, and `<form action>` is in fact allowed to be a function. Flagging both as false positives after merge if they persist; presumably these will be updated on their end eventually

PR #1028, opened 2024-12-14
remove new reddit platform handling

### LINKED ISSUE #1027: Remove support for new.reddit.com
"new Reddit" , `new.reddit.com`, has been removed. The `new.reddit.com` domain now redirects to `www.reddit.com` and the old/new preference in user settings now causes your www to be shreddit instead.

- ~~remove `new.reddit.com` from manifests~~ wait right, that never needed to be listed in manifests explicitly because we don't make web requests against it
- identify and remove code that's specific to new Reddit

eritbh:
Aside from the "open in new reddit" button that was already handled in #1025, there's not actually a whole lot of dead code we can remove by not supporting new Reddit. We still rely on the combination of TBListener + the oldreddit module for a bunch of UI stuff in old Reddit; eventually #940 will replace both with a new system and everything will be changed to use that instead, but that's not happening here. A couple features specific to new Reddit were removed, and the modbar `overlayFromBarRedesign` setting (which controls whether clicking queue buttons in the modbar on new reddit opens the queue as an overlay or acts as a normal link) now applies to shreddit instead.

Fixes #1027

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ All Toolbox subreddit settings and data are stored in subreddit wikis through ve
 [amo]: https://addons.mozilla.org/en-US/firefox/addon/reddit-moderator-toolbox/
 [edge-store]: https://microsoftedge.microsoft.com/addons/detail/moderator-toolbox-for-red/nolhfmoadpbgfeonfnjlcfmhhkmdlfcl
 [discord]: https://discord.gg/8fGCykQ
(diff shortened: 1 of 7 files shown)


eritbh:
branch updated to actually remove the lightbox setting and purge it from storage. Also cleaned up the related purges of removed settings - located them all together in a single `Promise.all`, and makes sure that they're purged regardless of the version we're updating from, since it's cheap to delete things even if they already don't exist

PR #1138, opened 2026-02-02
Fix OAuth token fetching

Littux-Dustux:
Fix OAuth token fetching method which will break on Feb 2, 2026

### DIFF
--- a/extension/data/background/handlers/webrequest.js
+++ b/extension/data/background/handlers/webrequest.js
@@ -2,63 +2,66 @@ import browser from 'webextension-polyfill';
 
 import {messageHandlers} from '../messageHandling';
 
+let cachedToken = null;
+
 /**
- * Retrieves the user's OAuth tokens from cookies.
+ * Retrieves an OAuth token from /svc/shreddit/token
  * @param {number} [tries=1] Number of tries to get the token (recursive)
- * @returns {Promise<Object>} An object with properties `accessToken`,
- * `refreshToken`, `scope`, and some others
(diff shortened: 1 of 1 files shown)


extension/data/background/handlers/webrequest.js
```diff
@@ -2,63 +2,66 @@ import browser from 'webextension-polyfill';
...
-    // http://stackoverflow.com/questions/20077487/chrome-extension-message-passing-response-not-sent
+    if (!cachedToken) {
+        cachedToken = await browser.storage.local.get('tb-accessToken') || {
+            accessToken: null,
+            expires: 0
+        }
+    }
+    if (cachedToken.expires > Date.now()) {
+        return cachedToken;
+    }
```
eritbh:
```suggestion
/**
 * Retrieves an OAuth token from /svc/shreddit/token
 * @param {number} [tries=1] Number of tries to get the token (recursive)
 * @returns {Promise<Object>} An object with properties `accessToken` and `expires`.
*/
async function getOAuthTokens (tries = 1) {
    const cachedToken = await browser.storage.local.get('tb-accessToken');
    if (cachedToken && cachedToken.expires > Date.now()) {
        return cachedToken;
    }
```

eritbh (changes requested):
Thanks for finding this method and submitting a fix - one minor code quality thing I'd like to address and then I can get this merged.

eritbh (approved):
Thanks again for contributing! Cleaned up the remaining issues myself, merging now

eritbh:
backported to 6.1 as b1143647

